PR #53, opened 2015-01-22
Apiman 238

ruromero:
Created the IConnectorInterceptor concept which can be set into the context in order to intercept remote connections.

### DIFF
--- a/gateway/test/src/test/resources/test-plan-data/simple/simple-connector-interceptor-policy/connector-interceptor/002-intercepted-response.resttest
+++ b/gateway/test/src/test/resources/test-plan-data/simple/simple-connector-interceptor-policy/connector-interceptor/002-intercepted-response.resttest
@@ -1,9 +1,7 @@
 PUT /gateway/SimpleConnectorInterceptorPolicyTest/connector-interceptor-policy/1.0/path/to/app/resource admin/admin
 Content-Type: application/json
+intercept: true
 
-{
-  "intercept" : "true"
-}
 ----
 200
 Content-Type: application/json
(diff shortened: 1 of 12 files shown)


gateway/test/src/test/resources/test-plans/simple/simple-connector-interceptor-policy-testPlan.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<testPlan xmlns="urn:io.apiman.test:2014:02:testPlan">
+
+  <testGroup name="Configure the Gateway">
+    <test name="Register Public Application">test-plan-data/simple/simple-connector-interceptor-policy/setup/001-publish-service.resttest</test>
+  </testGroup>
+
+  <testGroup name="Test the Connector Interceptor Policy through the Gateway">
+    <test name="Echo 1">test-plan-data/simple/simple-connector-interceptor-policy/connector-interceptor/001.resttest</test>
```
EricWittmann: Looks like you are invoking 001 but not 002-intercepted-response.  I think you need to add the latter into this list so that your feature actually gets tested!  :)

gateway/test/src/test/resources/test-plan-data/simple/simple-connector-interceptor-policy/connector-interceptor/002-intercepted-response.resttest
```diff
@@ -0,0 +1,13 @@
+PUT /gateway/SimpleConnectorInterceptorPolicyTest/connector-interceptor-policy/1.0/path/to/app/resource admin/admin
+Content-Type: application/json
+
+{
+  "intercept" : "true"
```
EricWittmann: Your test policy is looking for "intercept" in the request headers, not in the body of the request.  So I think this needs to be:

```
Content-Type: application/json
intercept: true
```

EricWittmann:
This looks really good, @ruromero - just a couple of inline comments.  If you could check those and then update the PR that'd be great.

PR #58, opened 2015-02-21
Implemented CachingPolicy APIMAN-219

ruromero:
Implementation of the Caching Policy.
The EchoServlet now includes a counter in the JSON response object and in a header.

Note: I have tried to test it but apparently I'm missing something in the configuration because the IDataStoreComponent is not found in runtime.

The changes includes a merge with the current master.


msavy:
It might be much cleaner if you try doing a rebase of your work on top of master (i.e. avoid the merge) as it's really hard to see your changes vs a whole load of other stuff that's been pulled in with the merge.

PR #59, opened 2015-02-23
Implementation of the Caching Policiy - APIMAN-219

ruromero:
Implementation of the Caching Policy.
The EchoServlet now includes a counter in the JSON response object and in a header.

Note: I have tried to test it but apparently I'm missing something in the configuration because the IDataStoreComponent is not found in runtime.

The changes includes a merge with the current master.


EricWittmann:
Thanks Ruben. I'll have a look at this and merge it in asap.

PR #65, opened 2015-03-09
fixing links on user-apps page to app-overview

KurtStam:


manager/ui/apiman-plugin/plugins/apiman/html/user-services.html
```diff
@@ -7,46 +7,17 @@
   </head>
 
   <body>
-    
+  <body>
+    <div ng-controller="Apiman.UserServicesController">
```
EricWittmann: I think the controller should instead be set on the "page controller" div (see below).  The navbar ng-include has its own controller.

KurtStam:
I guess so. I guess I'm in code you have not yet updated. Will the second controller still know about the current user when we hook that up?

manager/ui/apiman-plugin/plugins/apiman/html/user-services.html
```diff
@@ -7,46 +7,17 @@
...
-                    <a href="#">Logout</a>
-                  </li>
-                </ul>
-              </li>
-            </ul>
-          </div>
-        </div>
-      </div>
-    </div>
+    <div id="apiman-header" ng-include="'plugins/apiman/html/navbar.html'"></div>
```
EricWittmann: Before the navbar there should be a page loading progress indicator:

https://github.com/apiman/apiman/blob/master/manager/ui/apiman-plugin/plugins/apiman/html/org-plans.html#L11

manager/ui/apiman-plugin/plugins/apiman/html/user-services.html
```diff
@@ -7,46 +7,17 @@
...
 
     <!-- Page content -->
     <div class="page container" data-field="page">
```
EricWittmann: This should be where the controller is defined.  Also this needs ng-cloak and ng-show directives:

https://github.com/apiman/apiman/blob/master/manager/ui/apiman-plugin/plugins/apiman/html/org-plans.html#L13

KurtStam:
I'm leaving it to you to finish your update and will switch pages that are similar to the ones you have updated already.

EricWittmann:
OK.  Done and committed.

PR #67, opened 2015-03-10
adding consumer-org screen

KurtStam:
That manager/ui/apiman-plugin/defs.d.ts file is a pain.. the slashes are difft when you check in from windows..


EricWittmann:
Agreed - do we need it checked in?  Seems like it's generated by gulp...

PR #76, opened 2015-03-12
APIMAN-325 Add initialize method for metrics

msavy:
As discussed, uses reflection to call an initialize if it's defined, rather than an explicit interface.

### DIFF
--- a/gateway/engine/core/src/main/java/io/apiman/gateway/engine/impl/EngineImpl.java
+++ b/gateway/engine/core/src/main/java/io/apiman/gateway/engine/impl/EngineImpl.java
@@ -15,6 +15,9 @@
  */
 package io.apiman.gateway.engine.impl;
 
+import java.lang.reflect.InvocationTargetException;
+
+import io.apiman.common.util.ReflectionUtils;
 import io.apiman.gateway.engine.IComponentRegistry;
 import io.apiman.gateway.engine.IConnectorFactory;
 import io.apiman.gateway.engine.IEngine;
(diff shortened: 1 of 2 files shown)


msavy:
I'm open to this being called, for instance, validate instead of initialize if you think that makes more sense.

gateway/engine/core/src/main/java/io/apiman/gateway/engine/impl/EngineImpl.java
```diff
@@ -64,6 +67,7 @@ public EngineImpl(final IRegistry registry, final IPluginRegistry pluginRegistry
         
         policyFactory.setPluginRegistry(pluginRegistry);
         metrics.setComponentRegistry(componentRegistry);
+        initializeMetric(metrics);
```
EricWittmann: Can you call initialize on all of the engine's dependencies here instead of just the metrics one?  It'd be nice to have that option for everything.

msavy: Could you clarify what you mean by this? For instance, you'd like initialize called on each factory, or on each of the factory's entries (e.g. every policy, every component, etc)

EricWittmann: Just on the factories and registries.  Specifically, these:

https://github.com/apiman/apiman/blob/master/gateway/engine/core/src/main/java/io/apiman/gateway/engine/impl/EngineImpl.java#L39-L44

msavy:
I've rolled this all into a single commit doing what you requested.

PR #81, opened 2015-03-19
Header hashmap trim and stringutils

msavy:


msavy:
Please check out the impl to see if you're happy with it. Shouldn't really impact performance noticeably and I can get rid of trims from other places.

I think some of the errors users might have seen are due to simple issues like this (e.g. leaving erroneous whitespace before/after their token, etc)

PR #84, opened 2015-03-19
adding service-contracts

KurtStam:


KurtStam:
TBD - paging

KurtStam:
TBD:
- Hookup Gateway Test
- Hookup PW validation
- Where is the field to set the Type in the html?

EricWittmann:
There's no field currently because we only support one type.

PR #89, opened 2015-03-23
Timers for apiman

msavy:
Component interface plus implementations in Vert.x and Servlet.

Let me know if there are any gaps.


msavy:
Perhaps I rename and move the Servlet impl as the Default impl?  It should work fine on any platform as it's non-blocking (the scheduler fires up a thread for itself).

msavy:
I just realised the cancelAll isn't thread-safe, I'll fix it tomorrow.

PR #91, opened 2015-03-23
IBufferFactoryComponent Interface and implementations

msavy:
Thoughts:
 - Should we have a create method that accepts byte[] *without* any copying the array (or perhaps this should be the default behaviour).
 - Should we have a `relinquishBuffer`? Perhaps we can look at how Netty does its pooling/reuse. This is probably a strong candidate for reuse optimisations.


EricWittmann:
I noticed that this component isn't actually created yet, but happy to merge the PR anyway.

PR #95, opened 2015-03-28
Completed port of org-manage-members (including card flipping, search, drop-downs, etc).

msavy:
Things to note:
  - Some elements of styling seem to have changed slightly (there's now a search icon on the search box, and the text is no longer italicised). I think this is a result of @EricWittmann's apiman-search-box element as far as I can ascertain from a bit of CSS sleuthing. Still looks fine, IMO.

-  Sometimes I can see a slight flash when flipping the card in chrome. I will come back and do more experiments with ng-cloak (etc), but I haven't managed to solve that one yet...

Long slog and mildly traumatic, but finally my first significant angular commit: hurrah!


EricWittmann:
gj

PR #108, opened 2015-04-03
Description in-place editing

msavy:
Is flexible enough to re-use elsewhere if we decide to do that.

At the moment I haven't added styling, that's my next task!


msavy:
As you notice, I haven't done anything with the errors other than logging. I think we should have an error pane or something. We could implement that as a service/factory that you can push messages to (so multiple errors can be displayed). The first message arriving would unhide the element.

WDYT?

msavy:
OK, fixed that small issue

PR #110, opened 2015-04-10
Ui style fixes rebased 2

msavy:
Going to tidy this all up tomorrow and try to fix one or two tiny glitches, but putting it here so you can test it out. I've already fixed-up and squashed a bunch of stuff: went through lots of renditions to get something working nicely :o)...


msavy:
I think this is probably safe to merge now, and any tiny glitches we find can be fixed later

PR #125, opened 2015-05-12
switching to use the Jest client

KurtStam:


manager/test/api/src/main/java/io/apiman/manager/test/server/ManagerApiTestServer.java
```diff
@@ -190,18 +198,27 @@ protected void preStart() throws Exception {
...
+//                        .settings(ImmutableSettings.settingsBuilder().build()).node().start();
+//                client = node.client();
+//            } else {
+//                TransportClient tc = new TransportClient(ImmutableSettings.settingsBuilder()
+//                        .put("cluster.name", clusterName).build());
+//                tc.addTransportAddress(new InetSocketTransportAddress("localhost", 6600));
+//                client = tc;
+//            }
+
+        	String connectionUrl = "http://localhost:6500";
```
msavy: @EricWittmann @KurtStam For a standalone client like Jest should we have this as a configurable attribute (e.g. apiman.properties) rather than hard-coded?

KurtStam: @msavy - Yep absolutely :). Eric will apply some polish depending on which client (or both) you want to support going forward. The JestClient seems somewhat slower then the Transport Client ( 23.808 seconds for the api-tests vs 15 s with the Transport Client)

msavy: Ah okay, I see.

I'd expected the performance to be lower given additional overheads, but I'm admittedly surprised as to the degree :-(.

PR #127, opened 2015-05-21
For review: Implement mutual TLS/SSL

msavy:
Implement mutual TLS/SSL:
- RequiredAuth enum, capable of parsing endpoint properties to determine auth type. Mutual TLS, BASIC and DEFAULT (none).
- Use a given Services' endpoint config to determine authorization type
- Provide Auth settings via HttpConnector's apiman.properties section (or equivalent).
  This also provides additional functionality to standard one-way TLS.

  * trustStore - JVM defaults if null           | Store of trusted certificates (e.g. internal CAs)
  * trustStorePassword - none                   | The trust store password (if any)
  * keyStore - required                         | The client keystore
  * keyStorePassword - none                     | The keystore password (if any)
  * keyPassword - none                          | Password on the key itself (if any)
  * allowedProtocols - JVM defaults if null     | The allowed protocols (e.g. TLSv1.2)
  * allowedCiphers - JVM defaults if null       | The allowed cipher suites
  * allowAnyHost - false                        | Enable/disable hostname verification
  * allowSelfSigned - false                     | Automatically trust self-signed certs

  For standard TLS, available settings are (with defaults):

  * trustStore - JVM defaults if null
  * trustStorePassword - none         
  * allowedProtocols - JVM defaults if null
  * allowedCiphers - JVM defaults if null
  * allowAnyHost - false
  * allowSelfSigned - false
  * devMode - false (Use in development ONLY)   | Disables *all* checks, trusts everything

Available Cipher Suites & Protocols can be found at following resource:
  https://docs.oracle.com/javase/7/docs/technotes/guides/security/SunProviders.html

- By default non-mutual TLS is used (assuming we're doing TLS/SSL).
- Ensure fine granularity is allowed for cipher and protocol selections. Standard
  behaviour is to use global settings via the environment, but different endpoints
  may require different behaviour in the future, so it is facilitated through a
  custom factory.

I'm working on getting some robust testing now, but when we have our meeting perhaps we can go over some ideas I have for some nice tweaks you might be amenable to.


gateway/engine/core/src/main/java/io/apiman/gateway/engine/auth/RequiredAuthType.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * BASIC
+     */
+    BASIC("basic"),
+    /**
+     * Default (nothing)
+     */
+    DEFAULT;
+
+    public static final String ENDPOINT_PROPERTIES_KEY = "authType";
```
EricWittmann: Let's rename this to:

    authorization.type

In the future we can scope the additional authorization endpoint properties this way.  *For example*, basic auth might be:

    authorization.type: basic
    authorization.basic.credentials.username: sadmin
    authorization.basic.credentials.password: abc123!

msavy: :+1: sounds good

pom.xml
```diff
@@ -119,8 +119,8 @@
     <version.org.apache.felix.scr.annotations>1.7.0</version.org.apache.felix.scr.annotations>
     <version.org.apache.geronimo.specs.geronimo-jta_1.1_spec>1.1.1</version.org.apache.geronimo.specs.geronimo-jta_1.1_spec>
     <version.org.apache.geronimo.specs.geronimo-servlet_3.0_spec>1.0</version.org.apache.geronimo.specs.geronimo-servlet_3.0_spec>
-    <version.org.apache.httpcomponents.httpclient>4.4-beta1</version.org.apache.httpcomponents.httpclient>
-    <version.org.apache.httpcomponents.httpcore>4.4-beta1</version.org.apache.httpcomponents.httpcore>
+    <version.org.apache.httpcomponents.httpclient>4.4.1</version.org.apache.httpcomponents.httpclient>
+    <version.org.apache.httpcomponents.httpcore>4.4.1</version.org.apache.httpcomponents.httpcore>
```
EricWittmann: Nice - everything else still works I assume?  :)

msavy: All seems to be fine, yes.

msavy:
Am thinking that for the server-all overlay we should set allowAnyHost and allowSelfSigned to true, which should mimick existing functionality. Will do that tomorrow if there are no objections.

EricWittmann:
Sounds good to me.

gateway/platforms/servlet/src/main/java/io/apiman/gateway/platforms/servlet/connectors/HttpConnectorFactory.java
```diff
@@ -30,18 +35,27 @@
...
+    // 2WAY auth (i.e. mutual auth)
+    private SSLSessionStrategy mutualAuthSslStrategy;
+    private Map<String, String> config;
+
     /**
      * Constructor.
+     * @param config map of configuration options
      */
-    public HttpConnectorFactory() {
+    public HttpConnectorFactory(Map<String, String> config) {
```
msavy: Given that the config will be on IConnectorFactory (e.g. `apiman-gateway.components.IConnectorFactory.allowSelfSigned=true`) perhaps I should also namespace the keys to be something like apiman-gateway.components.IConnectorFactory.tls.allowSelfSigned=true

EricWittmann: +1

msavy:
If you check out https://github.com/msavy/apiman/tree/auth-refactor-rebase-4 I'm nearly there. Done some changes as I'd missed out a few things and also a few improvements, such as a flag to enable a development/debug SSL mode which is essentially an 'accept anything in all circumstances'. Appropriate dire warning emitted when instantiated.

Hit a few time consuming road bumps yesterday, which turned out to be a small issue that I overlooked.

Main task still to do is changing the 'namespacing' of the config variables, and putting them as constants in some sensible location (Maybe `RequiredAuthType`?).

msavy:
I think this is mergeable after a last review (will spend some time further squashing commits if there are too many)

PR #128, opened 2015-05-27
Add key aliases

msavy:
Add ability to define which key alias(es) will be used from the keystore

By default, if no alias list is provided, all keys are candidates to be used. If an alias list is provided, apiman will *only* attempt to use the listed keys. Invalid aliases will be ignored:

    apiman-gateway.components.IConnectorFactory.tls.keyAliases=<COMMA_SEPARATED_LIST>


EricWittmann:
Nice - good testing.

PR #139, opened 2015-06-05
fix for minification

msavy:


EricWittmann:
Thanks - this is exactly what I did but I forgot the quotes.  So the TS compiler yelled at me.  I was being dumb.

PR #141, opened 2015-06-09
APIMAN-455 Apiman favicon is now displaying

bgaisford:
Base tag specifies base URL for relative URLs.  Previous favicon was using a absolute URL.


EricWittmann:
Thanks!

PR #144, opened 2015-06-15
APIMAN-480 Add <finalName> tag to all poms using war packaging

bgaisford:
To aid in IDE/server integration, it is extremely helpful if projects that generate wars create war artifacts named to match the intended target context names.

### DIFF
--- a/gateway/platforms/war/wildfly8/api/pom.xml
+++ b/gateway/platforms/war/wildfly8/api/pom.xml
@@ -95,6 +95,7 @@
   </dependencies>
 
   <build>
+    <finalName>apiman-gateway-api</finalName>
     <plugins>
       <plugin>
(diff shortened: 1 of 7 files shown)


bgaisford:
GIve this a whirl Eric.  I don't see any negative side effects.  Is there any chance this could impact your release process?

EricWittmann:
I will give this a try asap, thanks.  I don't think it should be an issue.

bgaisford:
We could also define these final war names as properties in the root pom.  This would help with maintenance down the road if the context names were to change.  Up to you.

EricWittmann:
Thanks.  No need to pull the names up as properties I don't think.  We can do that later if there becomes a compelling reason for it. :)

PR #148, opened 2015-06-18
APIMAN-451 Entities updated to remove camelCase in database columns

bgaisford:
As part of the larger APIMAN-451 DDL generation task, it is necessary to make these updates to be more compatible across database engines.


bgaisford:
MIssed one. :(

PR #149, opened 2015-06-19
APIMAN-451 Initial liquibase DDL generation

bgaisford:
This is a work in progress.  I wanted to get this into source control so other may begin looking at it.  In its current state, the project generates DDLs for h2, mysql and postgresql based on latest JPA entities and static pre-population data.


bgaisford:
In the second commit I removed the postgres and mysql generation as those require database connection you may not have yet. :)

EricWittmann:
Thanks!  Will have a look as soon as I can.  

@msavy and @brmeyer  - if not too busy, can you guys also have a look and offer opinions?

bgaisford:
@EricWittmann @msavy @brmeyer - I'm working on a high level doc describing the current state of this effort, choices made so far, proposed workflows and next steps.  It may be worthwhile waiting to review the PR until doc is available.  I'll do my best to send out today, for sure by Monday.

bgaisford:
@EricWittmann apiman-distro-db-ddl may be a better project location.  I'm not married to the current location in the pull request.

EricWittmann:
It's fine.  Easy enough to change it later if need be. :)

distro/ddl/pom.xml
```diff
@@ -0,0 +1,175 @@
...
+    <!-- Third Party Dependencies -->
+    <dependency>
+      <groupId>org.hibernate.javax.persistence</groupId>
+      <artifactId>hibernate-jpa-2.1-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-entitymanager</artifactId>
+      <version>4.3.1.Final</version>
```
EricWittmann: All deps should be managed in the apiman/pom.xml root pom.

distro/ddl/src/main/db-setup/ReadMe.txt
```diff
@@ -0,0 +1,38 @@
+This directory contains setup instructions and scripts for required live databases
```
EricWittmann: Can you switch this to an .md file?

bgaisford: Yes, will do.  I've already done that on my side.  Will add an .md in the project root.

distro/ddl/pom.xml
bgaisford: I'll have experiment with this.  With regard to the apiman dependency, we don't want to inherit the whole apiman project classpath.  I ran into this issue before, the entity bean scan that we've configured for liquibase will discover other persistence.xml files and attempt to act on those and fail.  Goal here is to limit the classpath to include only entities.

EricWittmann: You're already inheriting the dependency management from the root - you should just need to remove the <version> from the dependency here.  In fact, the hibernate-manager dependency is already managed by the root, so really just remove the <version> tag here.

There's no need to do this for the liquibase plugin config - that can stay as you have it (very specialized to this project).

bgaisford: Will do.

bgaisford:
I'm think this is probably good for phase one.  DDL generation working well without live databases, indexes added, etc.  Next phase in a new PR we'll reconcile entity updates against offline snapshots using the new 3.4.0 offline support released by liquibase.

EricWittmann:
I'm going to review this hopefully tomorrow.  What I've seen so far looks good. :)

bgaisford:
Sounds great Eric, thanks!  I think a good sanity check will be to do an exhaustive test against a database populated with one of the generated DDLs just to be sure we haven't missed anything critical - I don't think we have though.

EricWittmann:
This looks nice and clean so far.  I will carve out some time to test against mysql as soon as I can.

PR #151, opened 2015-07-06
Locale workaround

msavy:


EricWittmann:
Can you put the locale restoration code in a try/finally block so it's guaranteed to be restored even if the test fails?

msavy:
They are in the @BeforeClass and @AfterClass sections, so should still run even when tests fail, right?

EricWittmann:
Sorry I must have looked at the diff to fast!  My bad.  Looks good

PR #156, opened 2015-07-10
APIMAN-452 Policy definitions now verify IPv4 + wildcard format

bgaisford:
Added a regular expression to match on IPv4 addresses + to also allow the '*' wildcard for octets within the IP address.


EricWittmann:
Thanks!

PR #157, opened 2015-07-11
APIMAN-458 Add meaning/purpose to the *Admin permissions

bgaisford:
Updated ActionResourceImpl to hook-up the appAdmin, svcAdmin and planAdmin roles for Register/Unregister, Publish/Retire and Lock operations as appropriate.  Expanded the rest publishing test plan to include testing role membership support against the *Admiin operations.  Looks good.


bgaisford:
I went through three implementations of the test plan for this issue.  In the end I sided with glomming on to the existing "publishing" test plan - I also refrained from renaming existing publishing test plan files.  I think it would be a mistake to copy the existing create/publish/retire test plan data to a new test plan as this just creates more maintenance and complicates matters.  Have a look at it Eric, open to suggestions for improvement.

EricWittmann:
This looks fantastic.  Definitely don't want a whole new test for this.  Nice job.

PR #173, opened 2015-08-13
Ensure reliable testing under Java8

msavy:
I discovered some glitches, which I've managed to resolve (thanks StackOverflow!)

Essentially, I am now properly exporting the test classes from the vertx3 gateway, which ensure they're available on the classpath when you run `mvn clean test -Dapiman.gateway-test.config=vertx3-mem`. By default they're not available, so they may/may not be on the classpath.

Also ensure that a few other things are reliably on the classpath (in-memory). See: https://github.com/apiman/apiman/compare/apiman:master...msavy:8fixes?expand=1#diff-0bc78dba935aa46d68cf49b963ef5a65R135

I've double checked and tests on java8 for the vert.x 3 port are always passing now.

### DIFF
--- a/gateway/platforms/vertx3/pom.xml
+++ b/gateway/platforms/vertx3/pom.xml
@@ -196,6 +196,19 @@
           </execution>
         </executions>
       </plugin>
+
+      <plugin>
+	<groupId>org.apache.maven.plugins</groupId>
+	<artifactId>maven-jar-plugin</artifactId>
+	<version>2.6</version>
+	<executions>
+	  <execution>
+	    <goals>
+	      <goal>test-jar</goal>
+	    </goals>
+	  </execution>
+	</executions>
+      </plugin>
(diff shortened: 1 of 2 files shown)


gateway/platforms/vertx3/pom.xml
```diff
@@ -196,6 +196,19 @@
           </execution>
         </executions>
       </plugin>
+
+      <plugin>
+	<groupId>org.apache.maven.plugins</groupId>
+	<artifactId>maven-jar-plugin</artifactId>
+	<version>2.6</version>
```
EricWittmann: versions of plugins should be managed at the root

msavy: That's a whoopsie. Just need to push the fix.

msavy:
Fixed that and weird indentation issue

PR #176, opened 2015-08-14
Simple travis file - will add more advanced bits later

msavy:
Bonus once we've enabled Travis is that all PRs will automatically be tested against my config. Pretty nice!


helio-frota:
+1

PR #181, opened 2015-08-18
Added skipIfFileExists flag to the ant get task

kunallimaye:
# Maven profile install-all-wildfly8
This profile by default downloads the wildfly binary from the net. Adding a check to the ant get task to check if we already have the binary and skip that task if the binary exists locally.
> skipexisting="true"

### DIFF
--- a/tools/server-all/pom.xml
+++ b/tools/server-all/pom.xml
@@ -1,218 +1,236 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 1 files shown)


msavy:
@EricWittmann AFAICT we already do this? Certainly seems to be the behaviour on my machine.

@kunallimaye Unfortunately your editor has reformatted and indented the XML so drastically I can't actually inspect the diff in a useful fashion. Every line shows as changed.

EricWittmann:
Yeah strange.  The ant docs suggest that skipexisting defaults to false but it seems to work for me anyway.  That said I'm happy to add the flag anyway.  

+1 on the formatting problems though.  Are there any other changes in there?

kunallimaye:
Nope no other changes. 
Happy to submit one with original formatting if required. Please let us know. 
--kl

--
Kunal Limaye

EricWittmann:
It's ok - that's a simple enough change that I just went ahead and made it:

https://github.com/apiman/apiman/commit/e970378d49a07c3e2a630ccffc43cf0eb67eb1ea

PR #187, opened 2015-08-21
Fix host problem in vertx3 port

msavy:


msavy:
Weirdly this didn't fail on my machine, but I probably didn't do a full clean install from the root.

The reason this is happening is because I added the eb-inmemory engine to vert.x 3, and it has introduced a cyclical dependency issue. I would like this to be available, though, because it's handy to have as a dumb test-it-out setup for blogs, etc.

I guess I might have to refactor and put some stuff in vert.x3-common (or whatever), which is a bit of a pain as it's nice having everything compact.

EricWittmann:
So should I merge them?  I guess not?

msavy:
No

msavy:
Will work to fix them before Monday

PR #197, opened 2015-08-26
[1.1.x] Cipher and protocol blacklisting

msavy:


msavy:
Haven't managed to reproduce usefully on my machine yet - seems to be that sometimes one exception appears, and other times another slightly different one (result seems to be correct).

PR #201, opened 2015-08-29
Entity updates to support NOT NULL java.util.Date columns

bgaisford:
Added @Temporal(TemporalType.TIMESTAMP) to all Date columns.  Also added @PrePersist and @PreUpdate to support NOT NULL Date columns within entities.  Hopfully this will address issues within Apiman relating to MySQL NOT NULL timestamps.


bgaisford:
@EricWittmann I just wanted to get this in a PR, not sure you'll use it.  We need to talk more about the MySQL TimeStamp/DateTime issue as I still don't know the particulars.  I was able to load the MySQL DDL without issue based on the liquibase changelogs prior to all your updates.  If it's a runtime issue you were seeing then perhaps you can fill me in further next week.

bgaisford:
This isn't required.  The Temporal annotations are probably a good idea in general, but better to wait until an issue arrises where needed.

PR #205, opened 2015-09-02
I want to override these method, so making them public

KurtStam:


EricWittmann:
Why can't you override them now?  Protected methods can be overridden in java.

msavy:
Any reason protected isn't good enough if you're overriding?

PR #209, opened 2015-09-03
APIMAN-556 Updates to correct liquibase DDL generation issues

bgaisford:
Added distro/ddl to the apiman build process
Updated ddl SNAPSHOT version to match project -> 1.2.0-SNAPSHOT
Upgraded liquibase to version 3.4.1
Changed ddl generation goals to execute during maven compile phase
Added global property to master.xml to treat MySQL clob type as longtext
Synchronized the changelogs with the entity generated changelogs prior to manual updates
Keeping manual LONGBLOB update in the changelogs - will need to document this for future JPA @Lob columns
Added defaultSchemaName to PostgreSQL updateSql goal - see comments in pom
Successfully loaded MySQL DDL
Successfully loaded PostgreSQL DDL

### DIFF
--- a/distro/ddl/pom.xml
+++ b/distro/ddl/pom.xml
@@ -161,7 +161,7 @@
               <goal>updateSQL</goal>
             </goals>
             <configuration>
-              <migrationSqlOutputFile>${basedir}/target/apiman_mysql.ddl</migrationSqlOutputFile>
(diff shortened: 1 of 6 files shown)


bgaisford:
@EricWittmann If you could load the distro/ddl generated MySQL DDL on your side and verify runtime against Apiman I think we're good to go on this.  Next step...add support to handle entity updates...I'll write up a new case in JIra.

EricWittmann:
Will do, thanks!

PR #214, opened 2015-09-07
[Do not merge] Move sonatype snapshot repo into parent pom temporarily

msavy:
@jcechace let me know if it works okay.


jcechace:
works fine.

msavy:
The reason I want to check this first with Eric is that we don't ever want to accidentally include snapshots in our releases (it's okay with the tech preview or whatever, but not a proper release).

msavy:
@EricWittmann You OK with this change? If so, should be safe to merge (I'm not aware that it'd cause any unwanted side-effects, but wanted to check with you first).

PR #215, opened 2015-09-08
More elegant method to deploy verticles & fix lack of https verticle

msavy:


msavy:
Looks OK now.

Aside: We should talk about how we can increase time-out periods to prevent spurious failures.

PR #218, opened 2015-09-10
Use newer converters

msavy:


msavy:
@jcechace This should fix the problems we were discussing yesterday, plus I've reintroduced code generation that was commented out before because it missed Map fields (Julien on the Vert.x team fixed that stuff for us, which was fantastic).

As far as I can tell, it works exactly as it should, but please let me know!

BTW, not sure if it's helpful, but I have some Vert.x-specific in-memory components which might be handy for testing (but you can only have 1 of each node type, as everything but the main registry is lazy loaded).

PR #224, opened 2015-09-24
re #675 Fix for Editable Description Directive

kahboom:
Changes:
- Fixes issue with editable description directive, where the changes were not being saved.

JIRA Issue: https://issues.jboss.org/browse/APIMAN-675

cc @EricWittmann

### DIFF
--- a/manager/ui/hawtio/plugins/api-manager/html/directives/editDescription.html
+++ b/manager/ui/hawtio/plugins/api-manager/html/directives/editDescription.html
@@ -1,5 +1,5 @@
 <div style="position: relative;" class="description">
-    <div editable-textarea="descr" e-ng-focus="focusOnDescription($event)" e-ng-change="changeOnDescription()"
+    <div editable-textarea="descr" onaftersave="saveDescription()" e-ng-focus="focusOnDescription($event)" e-ng-change="changeOnDescription()"
         e-class="autoExpand" ng-mouseover="descriptionMouseOver($event)" ng-mouseout="descriptionMouseOut($event)"
         id="descriptionWrapper">
             {{ descr || "&lt;" + defaultValue + "&gt;" }}
--- a/manager/ui/hawtio/plugins/api-manager/ts/directives.ts
+++ b/manager/ui/hawtio/plugins/api-manager/ts/directives.ts
@@ -439,11 +439,8 @@ module Apiman {
                     $scope.leftPosition = 0;
                     $scope.height = 60;
                     
-                    console.log('$scope.desc: ' + $scope.desc);
-
                     // If description is updated, call updateFunction.
                     $scope.$watch(function() {
-                        //console.log('Description: ' + $scope.org.description);
                         return $scope.descr;
                     },
                     function(new_value, old_value) {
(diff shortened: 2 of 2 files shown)


EricWittmann:
I think this is how I attempted to fix it as well.  However, I seem to recall that doing this introduced the undesirable behavior of invoking the callback in some cases even when the value hadn't actually changed.  Can you double check that?  The symptom would be that the UI will make REST calls to update the entity at inappropriate times (e.g. when first navigating to the Org page).

kahboom:
@EricWittmann - Ahh gotcha. Yeah, that's my fault since I don't have tests written for this. I'll take a look first thing in the morning.

Btw, do we have unit tests set up at all on the FE? If not I can work on this throughout the week.

EricWittmann:
Sadly no we have zero FE unit tests

kahboom:
No worries I'll work on that.

On Thursday, September 24, 2015, Eric Wittmann <notifications@github.com>
wrote:

> Sadly no we have zero FE unit tests
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apiman/apiman/pull/224#issuecomment-143097356>.
>


-- 
Sent from Gmail Mobile

msavy:
@EricWittmann @kahboom  +1 this is a huge area to work on (in addition to the funky mechanism for sending requests to the backend)

kahboom:
So, there were a few things I noticed could be improved, and this is a first step toward working on that. For instance, I think we should avoid `$watch` unless necessary, because as the application grows it could slow it down considerably. I noticed it was watching and executing the `$watch` callback during unintended moments, for instance on page load before the value was selected or changed. So, I removed `$watch` and added a method that the form will call only once it is submitted.

The only downside to this is that if the user submits the form and the value is exactly the same as the old value, it will still execute the callback. This can be corrected if you'd like, by storing the original value and then comparing it to the new value. Still, I think that would be considered an edge case since most users would normally just cancel the form.

--

@EricWittmann @msavy :
Regarding tests: Do you guys have an opinion on what unit testing tools to use? I normally use a setup similar to [this](http://www.smashingmagazine.com/2014/10/introduction-to-unit-testing-in-angularjs/), which is basically Karma for the test runner, Mocha for the unit testing framework, Sinon if we need mocks/stubs/spies, especially for testing service requests to the API, and Chai for assertions.

If it sounds like too much, we can skip Chai and just use Node.js's internal assert module, and we could probably skip Sinon and write some workarounds. Let me know what you think, if I'm getting too ahead of myself, if you prefer any other tools, etc.! Thanks :)

msavy:
Yeah I recall I spent ages trying to get exactly the behaviour we wanted, but it was harder than I anticipated (and evidently I still didn’t quite succeed). I think the compromise you suggested is reasonable.

I'll let Eric make the determinations, but Mocha seems kinda like Ruby's rspec/cucumber, which I think is great. I'm happy to use all the right tools to make our lives easier and things as robust as possible :-).

EricWittmann:
I don't have a toolchain preference with respect to unit testing the FE.  

Note: ideally whatever we use should be integrated into the maven build.

kahboom:
@EricWittmann: Okay np, I'll look into that. :+1:

PR #226, opened 2015-09-29
re #306 Fix for user not being prompted to save edits

kahboom:
Changes:
- Add a `$rootScope.isDirty` flag that is globally accessible within the app.
- Switch out references to `$scope.isDirty` across app to instead set new `$rootScope.isDirty` value.
- Add a check on request to change route, which checks the `$rootScope.isDirty` value.
- Add a prompt if the `$rootScope.isDirty` value is true, which warns the user about losing unsaved changes, and allows them to cancel to stay on the same page.

Right now the prompt is just your classic 1990's JavaScript confirmation window, but this can easily be improved if we decide to do so later on.

cc @EricWittmann @msavy


EricWittmann:
This looks great!

However, can you squash the commits down into a single commit?  That's not *always* a useful thing to do, but in this case the extra commits don't add clarity to the solution.

kahboom:
Closing, see https://github.com/apiman/apiman/pull/227

PR #229, opened 2015-09-30
implemented proper delete/re-add of plugins

EricWittmann:
...as well as upgrading of plugins to newer versions


manager/ui/hawtio/plugins/api-manager/ts/forms/edit-gateway.ts
```diff
@@ -89,6 +89,7 @@ module Apiman {
                 $scope.updateButton.state = 'in-progress';
                 var gateway = Gateway();
                 ApimanSvcs.update({ entityType: 'gateways', secondaryType: $scope.gateway.id }, gateway, function() {
+                    $rootScope.isDirty = false;
```
msavy: Is this a stowaway change? Doesn't seem related :)

EricWittmann: Yeah I slipped a bug fix in there too.  Good catch. :)

EricWittmann: @kahboom - this was the fix to the edit-gateway page.  The bug was that I was getting the "you haven't saved your changes yet!" dialog when clicking on the 'Save' button.  The reason is that the 'Save' button on that page results in a redirect of the user to a new page.  So the solution was to set the isDirty flag to false if the save worked, prior to doing the redirect. :)

kahboom: @EricWittmann - Thanks for tagging me on this. That was a good catch, I hope we don't have that issue in other areas. I'll do some additional testing later today. :)

EricWittmann: I think that's the only place.  But only because many of the other "edit"
pages don't implement isDirty functionality at all.  :)  We can add that to
the ever growing list of improvements to make.

PR #232, opened 2015-10-01
re #707 (partial) - Rm Hawt.io Nav & Route Handling

kahboom:
Changes:
- Remove Hawt.io route handling and tabs.
- Add route handling directly within application.
- Remove references to Hawt navigation from the main app.
- Dashboard is now `/`, instead of `/dash`.

To Do:
- I left the `api-manager` prefix for now, and will have to work on this separately (the goal is to remove it completely). All of the routes worked without issues, except for when you attempted to use a service, which likely has to do with the way they're are set up right now. I can work on this in a separate feature branch.
- The `hawtio-core`, `hawtio-core-dts`, and `hawtio-utilities` modules are still there, because it was getting quite complicated and time-consuming to remove them. I think I managed to make good progress toward this, and left the work in this feature branch (https://github.com/kahboom/apiman/tree/707-remove-hawtio) in case anyone wants to take a look. I was still getting errors in the browser console and the app was not loading properly. This probably has to do with my lack of familiarity with TypeScript (and TS interfaces).

NOTE: These changes mean that the dashboard will now be located at `http://localhost:2772/api-manager/`, with the hopes of becoming `http://localhost:2772` once I finish that on the other feature branch.

cc: @EricWittmann  Let me know if you want me to squash my commits. This came from a feature branch so should be a little easier this time. :)


EricWittmann:
@kahboom - yeah please do a squash down to a single commit if you don't mind too much.  I appreciate your local approach of commit-early, commit-often, though.  But when you do a PR I think squashing down to a single commit per PR might be better. :)

msavy:
I think something might have gone a bit funky here with a rebase/merge or something,  hence one of your commits appears randomly in the middle here.

kahboom:
@EricWittmann , You got it :+1: 

@msavy , I tend to have that problem. It's something I'm doing, but can't figure out what it is. What I did was run `git pull upstream master` from my feature branch, maybe this process would fix that: `git checkout master`, `git pull upstream master`, `git checkout <feature branch>`, `git rebase master`.

EricWittmann:
Ah yes - you should def. not do a pull from upstream master onto your feature branch.  The rebase option is your best bet.

msavy:
I've fixed this up and will give you a branch in a sec, but please triple check it because I had to do some conflict resolution as I think there's a very high chance I didn't get it quite right.

kahboom:
Closing. See https://github.com/apiman/apiman/pull/234.

PR #237, opened 2015-10-08
Change error output to align with servlet behaviour/new tests

msavy:
This makes the tests green.

### DIFF
--- a/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/http/HttpExecutor.java
+++ b/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/http/HttpExecutor.java
@@ -15,12 +15,8 @@
  */
 package io.apiman.gateway.platforms.vertx3.http;
 
-import java.util.Map.Entry;
-import java.util.UUID;
-
-import org.apache.commons.lang3.exception.ExceptionUtils;
-
 import io.apiman.common.util.MediaType;
+import io.apiman.gateway.engine.beans.EngineErrorResponse;
 import io.apiman.gateway.platforms.vertx3.common.config.VertxEngineConfig;
 import io.apiman.gateway.platforms.vertx3.io.VertxApimanBuffer;
 import io.apiman.gateway.platforms.vertx3.io.VertxPolicyFailure;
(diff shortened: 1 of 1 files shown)


msavy:
Sorry, I hadn't quite done this before the other one was merged. I'd normally have squashed it all together.

msavy:
Looks like I might have spoken too soon and perhaps some issues have been introduced when using ES + Vert.x 3 :(. Will investigate soon.

gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/http/HttpExecutor.java
```diff
@@ -166,7 +164,9 @@ private void setError(Throwable error) {
         response.setStatusCode(HttpResponseStatus.INTERNAL_SERVER_ERROR.code());
         response.setStatusMessage(HttpResponseStatus.INTERNAL_SERVER_ERROR.reasonPhrase());
         response.headers().add("X-Exception", String.valueOf(error.getMessage())); //$NON-NLS-1$
-        response.write(ExceptionUtils.getStackTrace(error));
+        response.headers().add(HttpHeaders.CONTENT_TYPE,  MediaType.APPLICATION_JSON);
+        response.write(String.format("{\"responseCode\":%d, \"message\":\"%s\"}",
+                HttpResponseStatus.INTERNAL_SERVER_ERROR.code(), error.getMessage()));
```
EricWittmann: I think you need to use jackson here - this code will produce an invalid response payload if the message contains quotes.  It's also missing the stack trace, which is supposed to be part of the response (but isn't tested due to the inconsistency of stack traces).  See here:

https://github.com/apiman/apiman/blob/master/gateway/platforms/servlet/src/main/java/io/apiman/gateway/platforms/servlet/GatewayServlet.java#L373-L389

msavy: I was printing the stack trace before, but it was failing the tests because of it. I swear I'd done this before but somehow I must have lost it/reverted it. 

Anyway, will fix.

EricWittmann:
Yeah I changed how it works recently.  It used to be that the stack trace was returned as the body of the response.  But I changed it so that a JSON payload is returned instead.  That's why the test started failing.  I thought I mentioned to you that I made that change and that it would break vert.x but maybe I forgot.  :(

Again, see the code I linked above for how it should work now.  :)

msavy:
Have force pushed fixed version, will see if it passes CI

gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/http/HttpExecutor.java
```diff
@@ -166,7 +166,14 @@ private void setError(Throwable error) {
         response.setStatusCode(HttpResponseStatus.INTERNAL_SERVER_ERROR.code());
         response.setStatusMessage(HttpResponseStatus.INTERNAL_SERVER_ERROR.reasonPhrase());
         response.headers().add("X-Exception", String.valueOf(error.getMessage())); //$NON-NLS-1$
```
EricWittmann: Can you make this X-Gateway-Error instead of X-Exception?

PR #238, opened 2015-10-08
Add timeout option for jest client, and double default T/O in many places - useful in CI scenarios

msavy:
Not sure whether this was worth adding as an extra config option in various places, so I just hard-coded a larger default in those areas.

Where it was easy I added an extra option to make it configurable (EsClientFactory).

Before merging it might be worth putting this one through CI a few times to see whether it trips anymore.


EricWittmann:
Travis is still failing in vertx, but the rest fail.  The code looks good to me so I'm merging.

PR #239, opened 2015-10-09
removed the idm storage interface, merging its methods into IStorage …

EricWittmann:
…and IStorageQuery


msavy:
I think you'll need to terminate the CI on this one. Somehow it's been running for 14 hours. That's pretty impressive.

PR #241, opened 2015-10-13
re #714 #485 Disable Fields if Service is Published

kahboom:
Changes:
- Added public method to policies file (parent controller), to allow check for entity status.
- Added `ng-disabled` directive to each field for each service tab, if said check returns a status other than 'Ready'.
- To handle custom forms, which are currently generated with JSON-Editor: Added a few statements to the `JsonSchemaPolicyConfigFormController`, which checks the entity status and disables the entire custom form if the status is anything other than 'Ready'. In the future, we can disable individual fields, but for the moment it seems to be a little bit more trouble than it's worth, so, for now, disabling the entire form seems to be the way to go.

Reference
- ["Page is dirty" warning sometimes happen when the page isn't dirty](https://issues.jboss.org/browse/APIMAN-714)
- [Grey out disabled fields](https://issues.jboss.org/browse/APIMAN-485)

cc @EricWittmann

### DIFF
--- a/manager/ui/hawtio/plugins/api-manager/html/forms/edit-policy.html
+++ b/manager/ui/hawtio/plugins/api-manager/html/forms/edit-policy.html
@@ -2,19 +2,31 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0"></meta>
+    <meta name="viewport"
+          content="width=device-width, initial-scale=1.0"></meta>
   </head>
 
   <body>
     
     <div ng-include="headerInclude"></div>
-    <div ng-controller="Apiman.EditPolicyController" id="form-page" class="container apiman-edit-policy apiman-entity-new page" data-field="page" ng-cloak="" ng-show="pageState == 'loaded'">
+    <div ng-controller="Apiman.EditPolicyController"
+         id="form-page"
+         class="container apiman-edit-policy apiman-entity-new page"
+         data-field="page"
+         ng-cloak=""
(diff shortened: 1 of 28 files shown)


EricWittmann:
Unfortunately you need to test for an entity state of "Created" as well as "Ready".  An entity is editable when in either of those states.  The difference between "Created" and "Ready" is that the latter indicates that the entity is "complete" and can be published/registered/locked.

I would recommend adding another function called isEntityEditable() to capture the comparison.  Such a function could be defined wherever getEntityStatus() is defined.

kahboom:
@EricWittmann - You got it; should be a quick update.

kahboom:
@EricWittmann - This is ready to be reviewed again. It was a bit more complicated than I thought; I had to add the method to scope of each policy controller due to the way I changed the way it checks whether or not to disable a field (adding that check method to compare strings vs checking strings directly in the DOM), which I didn't have to do before. But this way is much cleaner as it separates logic from the DOM. In the future, refactoring the app in specific areas would reduce the amount of code duplication.

Thanks! :)

kahboom:
Note: I was just able to reproduce issue [714](https://issues.jboss.org/browse/APIMAN-714), even after disabling fields. So this PR will only resolve issue [485](https://issues.jboss.org/browse/APIMAN-485). Working on resolving 714 now.

kahboom:
@EricWittmann - Ready for review again. Re-tested using both set of steps to reproduce, and this seems to have fixed it. Let me know if you still get any unexpected warnings about unsaved fields.

EricWittmann:
Looks good!

PR #243, opened 2015-10-14
adding swagger 2.0 swagger.json/yaml endpoint

KurtStam:


msavy:
@EricWittmann CI failure seems to be the `SimpleEchoTest.after:43 expected:<9> but was:<8>` thing again. 

@KurtStam As I understand the code, this should essentially auto-generate a swagger 2 spec, which will reside at `/api` of the manager microservice? Correct?

KurtStam:
Yes, it adds /swagger.json to the root context, for management it adds apiman/swagger.json and for the gateway api/swagger.json.

gateway/platforms/war/micro/pom.xml
```diff
@@ -159,6 +159,18 @@
       <artifactId>jboss-servlet-api_3.0_spec</artifactId>
     </dependency>
     <dependency>
+      <groupId>io.swagger</groupId>
+      <artifactId>swagger-jaxrs</artifactId>
+      <optional>true</optional>
```
EricWittmann: Why is this marked as optional?

KurtStam: Was considering making it an optional feature but changed my mind later - feel free to remove it.

EricWittmann: Thanks.  Already done.

PR #244, opened 2015-10-15
re #707 Remove Hawtio deps, add custom TSD & Gulp task for third-party TSD

kahboom:
Changes:
- Remove Hawtio-related dependencies: hawtio-core, hawtio-core-navigation, hawtio-utilities, hawtio-core-dts
- Bootstrap Angular app manually.
- Load app configurations manually.
- Update README with new setup instructions.
- Add new `setup` and `tsd` Gulp tasks to handle TS definitions that were previously handled by Hawtio.
- Add convenience npm script to run TSD and NPM installations.
- Update gitignore to not track IDE `.idea` and TS `typings` directories.
- Add missing dependencies to Browserify entry file.
- Other minor cleanup.

Ref: https://issues.jboss.org/browse/APIMAN-707

@EricWittmann @msavy - Once this gets merged in, you'll likely need to run `npm run setup` to reinstall NPM modules and to install new TSD packages. With dependency changes like these I always recommend wiping out npm modules completely and cleaning the cache prior to installing, like this:

`rm -rf node_modules/`
`npm cache clean`

Also, @EricWittmann, I wanted to move the app from `/ui/hawtio` to `/ui`, as we had discussed, but am not totally sure how to approach this, as there seems to be a `pom.xml` file for each directory. Maybe we can make that another task? I see Hawtio being referenced in each, and I removed one of the references. Please let me know if that will affect anything even though we are not using Hawtio anymore. Thanks!


kahboom:
Side note: I'm not happy with initial lag time on the very first page load (once it's cached, it's fast), but that's because we are manually bootstrapping the app (Hawtio was doing this as well, just through their library). Automatic bootstrapping would make it much faster on the first page load.

The downside is that we would have to move the content in the JSON files (`config.json` and `translations.json`) to instead be Angular constants. I personally like JSON better than Angular constants, but I also hate making an HTTP request to something the client should have access to from the start. Also, if the page load is a priority that would probably be the best workaround. Let me know what you guys think.

Some examples:
- http://twofuckingdevelopers.com/2014/06/angularjs-best-practices-001-constants/
- Facilitated with a Gulp preprocessor that injects it automatically as a constant module: http://stackoverflow.com/questions/25988293/how-to-read-the-json-config-file-inside-angular-module

kahboom:
Btw I noticed an error in the browser console that wasn't there before, maybe hold off on merging: `Uncaught SyntaxError: Unexpected token <`

It doesn't seem to affect that functionality as far as the very basic tests I've run, but it may drive me insane.

kahboom:
@EricWittmann - Fixed the browser console error mentioned above; it was due to a Hawtio IE shim I forgot to remove. Need to look into whether or not I will need an IE shim with the new TS setup. Anyway this is ready for review again. No rush!

PR #245, opened 2015-10-16
Import export support

EricWittmann:
Thanks marc!


msavy:
You took this a lot further than I had it, but I think I have a better understanding of the data layer now, so that's a big plus.

PR #261, opened 2015-11-03
Caching to hopefully avoid npm issues

msavy:


msavy:
I think the cache doesn't seem to kicked in yet (or it isn't working as expected).

@EricWittmann That being said, did we remember to set the TSD_GITHUB_TOKEN in travis' config page? That should theoretically fix this issue.

EricWittmann:
Yes - the TSD_GITHUB_TOKEN is set.

image

msavy:
Should be safe to merge, hopefully after a successful set of builds the cache will  kick in

PR #265, opened 2015-11-05
Bumping KC version to 1.6.1.Final

jcechace:
The old version of KC (1.2.0.Final) contains a bug which prevents configuration reload via cli

Running the following command right after starting a fresh distribution of apiman will crash it.

```
./bin/jboss-cli.sh -c ":reload"
```

### DIFF
--- a/manager/ui/hawtio/wildfly8/src/main/java/io/apiman/manager/ui/server/wildfly8/KeyCloakBearerTokenGenerator.java
+++ b/manager/ui/hawtio/wildfly8/src/main/java/io/apiman/manager/ui/server/wildfly8/KeyCloakBearerTokenGenerator.java
@@ -21,7 +21,7 @@ import io.apiman.manager.ui.server.beans.BearerTokenCredentialsBean;
 import javax.servlet.http.HttpServletRequest;
 
 import org.keycloak.KeycloakSecurityContext;
-import org.keycloak.util.Time;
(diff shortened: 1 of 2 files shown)


msavy:
Should also bump the version in apiman plugins - there's a possibility something in there might need moving around, too.

msavy:
Quick check shows that there are indeed a few refactors. I'll check that out and fix tomorrow morning, or you can do it if you prefer @jcechace .

msavy:
See https://github.com/apiman/apiman-plugins/pull/41

EricWittmann:
Did you guys check that the distro still works?  I'm guessing it does not, since KC 1.6.1 doesn't support wildfly 8.  So in order to accept this, we also need to upgrade to whatever version of wildfly KC supports.  We also need to make sure that the EAP 6.4 distro still works.

Also, the documentation and web site will (eventually) need to reflect the new WF version.

jcechace:
Yop, you are right -- I actually tried that in a hurry and wanted to know what Trevis CI will say about it -- without realising that these tests wont cover the distro. 

I will check tomorrow what is the latest working version  which does not suffer from the reload issue. 

This can be obviously close, sorry  :)

EricWittmann:
I'll close for now - but if you're interested in doing the full upgrade to KC 1.6.1 and WF 9.x I'm 100% in favor of it.  :)  I think we even have a jira or two about that...

PR #268, opened 2015-11-11
vertx with query params

msavy:
So, my solution here allows things to done cleanly with PreparedStatement under the covers - however, as you rightly pointed out the `execute` method doesn't have parameters for some reason. I've worked around that by actually executing a `updateWithParams`.

I'll make a PR upstream to see if they'll accept an addition of `executeWithParams`. Should be easy :).


EricWittmann:
This looks great!  I didn't see the WithParams variants - but TBH I didn't really look either. :)

Before merging, can you update a couple of your unit tests to use params?

msavy:
Roger

PR #286, opened 2015-11-30
Add OSGI Karaf branch

cmoulliard:
- Add missing apache maven felix plugin 
- Change packaging to bundle
- Include web.xml file to run jetty standalone

### DIFF
--- a/manager/api/micro/pom.xml
+++ b/manager/api/micro/pom.xml
@@ -7,6 +7,7 @@
     <version>1.2.0-SNAPSHOT</version>
   </parent>
   <artifactId>apiman-manager-api-micro</artifactId>
+  <packaging>bundle</packaging>
   <name>apiman-manager-api-micro</name>
 
   <properties>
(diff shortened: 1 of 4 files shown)


EricWittmann:
No need for a branch of these changes.  They are all non-destructive - so i'm going to simply merge them onto master.

manager/api/micro/pom.xml
```diff
@@ -238,5 +239,31 @@
...
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <extensions>true</extensions>
+        <configuration>
+          <instructions>
+          <Import-Package>
+            javax.servlet;version="[2.6,4)",
+            javax.servlet.http;version="[2.6,4)",
+            org.eclipse.jetty.*;version="[8.0,10)",
```
msavy: @EricWittmann @cmoulliard is there standard to way to track those version numbers so that if we alter ours (in the main apiman pom) that these will be reflected in the OSGi imports?

EricWittmann: I think the ranges specified (for these packages) is good enough.  It's only going to be a problem for major version upgrades of core technologies.  So e.g. when we want to move to servlet 5 or something.  The only one that might be an issue is swagger, since the upper bound is 2.0 and that's currently the version we need.

To answer the question - I don't know. :)

cmoulliard: We can generate the range using some magic symbols --> https://github.com/apache/camel/blob/master/parent/pom.xml#L562-L564

msavy: That seems a decent idea. Just use the Maven version properties directly. @EricWittmann WDYT?

EricWittmann: Sounds good to me.

cmoulliard: Agree.

PR #296, opened 2015-12-14
Async ldap component (with pools)

msavy:
I have significantly improved this since last PR! I should write a blog at some point to show off the new functionality, as you can do quite a bit with it.

Includes tests, servlet impl, vert.x 3 impl, pool, improved interfaces.


msavy:
It passes - https://travis-ci.org/msavy/apiman/builds/96846004

PR #292, opened 2015-12-02
re #803 Switch from bootstrap-select to UI-select

kahboom:
Changes:
- Switch to UI-select for the following views: Service Implementation tab, Service Definition tab, Service Plans tab, New Contract (&modal)
- Re-add validation for some UI-select dropdowns that were misbehaving
- Removed some expensive `$watch`'s but there are still many improvements that need to be made in ongoing PRs.
- Style for consistency on respective pages
- Fix an unrelated console error for User Profile page that I noticed a while ago
- Moved the `UserProfile` controller to the `/ts/user` dir, with the rest of the user stuff.
- Other minor cleanup for unused functionality or markup

The dropdowns should look very similar, with the exception of the ability to search for items. I've disabled this feature for dropdowns where it wouldn't make much sense, such as for versions. See screenshots below:

screen shot 2015-12-02 at 12 05 16 pm

screen shot 2015-12-02 at 12 06 24 pm

JIRA: https://issues.jboss.org/browse/APIMAN-803

We are still missing something like 20-22 views. The hard part was re-adding validation for a stubborn dropdown that was misbehaving (specifically the Gateway dropdown), and also removing some of the `$watch`'s. There are still a few improvements that need to be made, but since this is lower priority it will have to be done incrementally over several PRs.

cc @EricWittmann - Sorry for the huge PR; I'll try to keep it shorter next time.

PR #301, opened 2015-12-17
re #803 UI Select

kahboom:
This basically just uses the `master` branch with the renaming changes and adds my Select2 > UI-Select changes on top of it. I had to merge the changes manually, so @jcechace I would really like it if you could run your tests against this fork and branch instead of the other.

JIRA: https://issues.jboss.org/browse/APIMAN-803
Other PR (outdated now): https://github.com/apiman/apiman/pull/292

cc  @EricWittmann


EricWittmann:
This looks great!  I'll wait for @jcechace to run some tests before I merge, unless that takes too long. ;)

jcechace:
I will be able to run them later today (currently I do not have access to vpn).  I will let you know the result at the latest tomorrow.

EricWittmann:
Wow that's awesome, thanks!

jcechace:
There seem to be some issues when creating contracts. 

1.  When directly accessing ```/apimanui/api-manager/new-contract``` and filling in the form, the "redirect" after hitting create ends with 404 under some circumstances
2. Client App version which are already registered are also displayed in the select. Not sure whether this is intentional as part of the "re-register" functionality (but it sure is confusing). 

It also seems that some elements on various places are slightly different. Jarek will have a look at that on Monday and us know.

kasparjarek:
ad 1. This happend when user change ClientApp. The POST request to create contract is always send to the Client which was selected when the page had loaded.

kahboom:
@kasparjarek - I was able to replicate this- POST request uses the Client App that was automatically selected on page load as opposed to the one manually selected after.

@jcechace - Is this on WF or Jetty? I'm attempting to replicate the first issue but haven't been able to.

As for the second issue, I asked @EricWittmann about this and yes, it is intentional. Changes are allowed to be made to registered clients, because there is a "re-register" action now.

kahboom:
@jcechace - Actually, just replicated it, except I got a 409 error instead of 404. See screeenshot below. Looking in this now. Thanks!

screen shot 2015-12-21 at 10 10 06 am

kasparjarek:
@kahboom I actually specified the first issue from @jcechace. I am sorry if it wasn't clear.

kahboom:
Ahh, thanks @kasparjarek !

jcechace:
@kahboom @EricWittmann  As for the second issue -- in that case we should display a sort of "reminder" that the changes won't take immediate effect.

EricWittmann:
After creating a contract we redirect to the client details UI page - on
that page we can add something more obvious to indicate that the client has
changed since it was last published.  Could you add a jira for that?

jcechace:
@EricWittmann  https://issues.jboss.org/browse/APIMAN-856

Though I still think that users should be able to distinguish already registered version when creating a contract (or performing any other operation really) as it will make you think whether you really want to do changes to already registered client.

EricWittmann:
That's a fair point - maybe another jira then to enhance the New Contract
page with some sort of indicator showing the user that the client app they
selected has already been registered?

jcechace:
@EricWittmann https://issues.jboss.org/browse/APIMAN-857

kahboom:
@jcechace - Thanks for creating those jiras, I appreciate it. Would you mind re-testing? I just pushed up a fix for the issue where the incorrect app is used to create a new contract, and also a fix for another bug I came across where the submit button wouldn't always get enabled.

jcechace:
@kahboom The issue should be fixed. 

However  there seems to be some delay in rest layer before the related endpoint starts returning the newly created contract.  Although it could be just a glitch in our tests (not able to reproduce manually). I will have to investigate further when I'm back from PTO -- this should not prevent merging.

PR #304, opened 2015-12-21
Fix APIMAN-854

msavy:
@kasparjarek can you please test this out?


kasparjarek:
@msavy it still returns ClientNotFoundException

PR #310, opened 2015-12-23
#861 Fix for inability to save API definition

kahboom:
Since this is a blocking issue I just decided to revert the addition of UI-Select for this page.

I may be getting a little overeager here in trying to switch to UI-Select, but as I'm diving in, this switch is more problematic for some views than others. It seems like there are many `$watch`'s that are also integrated with other functionality, such as dirty checking and sometimes state setting.

While I'd prefer to have a cleaner implementation of these features, I know it's certainly not ideal nor efficient to rewrite these on the go, especially without testing in place. I think we should still switch to UI-Select, but it may be a while before we can transition completely.

JIRA: https://issues.jboss.org/browse/APIMAN-861

cc @EricWittmann

### DIFF
--- a/manager/ui/hawtio/plugins/api-manager/html/api/api-def.html
+++ b/manager/ui/hawtio/plugins/api-manager/html/api/api-def.html
@@ -43,7 +43,7 @@
               <span class="clearfix"
                     style="margin-top:10px; margin-bottom: 10px"
                     apiman-i18n-key="api-definition-label">API Definition:</span>
-              <select apiman-i18n-key="api-def.choose-type" title="Choose a type..." apiman-select-picker="" ng-model="selectedDefinitionType" class="selectpicker" ng-options="type.label for type in typeOptions track by type.value">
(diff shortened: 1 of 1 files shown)


manager/ui/hawtio/plugins/api-manager/html/api/api-def.html
```diff
@@ -43,17 +43,8 @@
...
-                         style="width: 30%; max-width: 300px;">
-                <ui-select-match placeholder="Choose a type..."
-                                 apiman-i18n-key="api-def.choose-type">
-                  <span ng-bind="$select.selected.label"></span>
-                </ui-select-match>
-                <ui-select-choices repeat="type in (typeOptions | filter: $select.search) track by type.value">
-                  <span ng-bind="type.label"></span>
-                </ui-select-choices>
-              </ui-select>
+              <select apiman-i18n-key="api-def.choose-type" title="Choose a type..." apiman-select-picker="" ng-model="selectedDefinitionType" class="selectpicker" ng-options="type.label for type in typeOptions track by type.value">
```
EricWittmann: I think this <select> box needs the ng-disabled="isEntityDisabled()" attribute included on it.

kahboom:
@EricWittmann - Done. Added `ng-disabled="isEntityDisabled()"` to `select`.

PR #314, opened 2015-12-30
Fixes inverted cache expiry logic.

outofcoffee:
The cache expiry logic in `InMemoryCacheStoreComponent.get()` is incorrect. Instead of checking that the expiry is in the past (as it does within the `getBinary()` method), it checks whether it is in the future.

This results in all cache entries retrieved via the `get()` method being considered immediately 'expired' and being removed from the cache.

This PR modifies the logic in `get()` to match that of `getBinary()` to fix the behaviour.


msavy:
Nice catch!

EricWittmann:
Thanks!

PR #323, opened 2016-01-12
Exit immediately if policy doesn't exist

msavy:


EricWittmann:
Is something similar needed for the ES registry impl?

msavy:
Looks like it’s okay from what I can see: https://github.com/apiman/apiman/blob/master/manager/api/es/src/main/java/io/apiman/manager/api/es/EsStorage.java#L864-L869 <https://github.com/apiman/apiman/blob/master/manager/api/es/src/main/java/io/apiman/manager/api/es/EsStorage.java#L864-L869>

PR #327, opened 2016-01-16
Fix Java 7 compile issue on 1.1.x

msavy:


msavy:
1.1.x can't be compiled on java 7 without this patch.

PR #330, opened 2016-01-20
Feature: When creating a contract with a custom "apiKey" so we can override original key generation.

ade90036:
Hi Eric, 

i've extended the Apiman REST call which creates a contract. Now it also takes an optional parameter which allows the creation of a contract with custom "apiKey". I thought it would also be useful to others. 

I've installed the code formatter but intelliJ had an opinion of its own, so you will see come commits reverting the file and manually apply the changes.

I've also extended a current test *.resttet to test this.

Maybe it would be work having a small section on the README, specifying how the test works, it took me a while to figure it out. (especially the resttest). 

Thanks

Andrea

### DIFF
--- a/manager/api/beans/src/main/java/io/apiman/manager/api/beans/contracts/NewContractBean.java
+++ b/manager/api/beans/src/main/java/io/apiman/manager/api/beans/contracts/NewContractBean.java
@@ -31,11 +31,16 @@ public class NewContractBean implements Serializable {
     private String apiVersion;
 
     private String planId;
+    private String apiKey;
 
     /**
      * Constructor.
      */
     public NewContractBean() {
+    }
+
+    public static long getSerialVersionUID() {
+        return serialVersionUID;
     }
 
     /**
@@ -95,6 +100,20 @@ public class NewContractBean implements Serializable {
     }
 
     /**
+     * @return the contractKey
+     */
+    public String getApiKey() {
+        return apiKey;
+    }
+
+    /**
+     * @param apiKey to set
+     */
+    public void setApiKey(String apiKey) {
+        this.apiKey = apiKey;
+    }
+
+    /**
      * @see java.lang.Object#hashCode()
      */
     @Override
@@ -105,6 +124,7 @@ public class NewContractBean implements Serializable {
         result = prime * result + ((apiId == null) ? 0 : apiId.hashCode());
         result = prime * result + ((apiOrgId == null) ? 0 : apiOrgId.hashCode());
         result = prime * result + ((apiVersion == null) ? 0 : apiVersion.hashCode());
+        result = prime * result + ((apiKey == null) ? 0 : apiKey.hashCode());
         return result;
     }
 
@@ -140,6 +160,11 @@ public class NewContractBean implements Serializable {
                 return false;
         } else if (!apiVersion.equals(other.apiVersion))
             return false;
+        if (apiKey == null) {
+            if (other.apiKey != null)
+                return false;
+        } else if (!apiKey.equals(other.apiKey))
+            return false;
         return true;
     }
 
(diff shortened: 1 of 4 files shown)


EricWittmann:
This looks good, thanks @ade90036 - but before I merge the changes do you think you could squash the commits into a single commit?  That will make it a bit easier to see what changed (in the future, should we need to).

You can squash the commits by doing this:

```
git rebase -i HEAD~9
```

In this case the "9" is the number of commits you want to squash.  This will cause a 'vi' editor window to be displayed containing the list of commits at the top of the file. Each line will start with word "pick". With the exception of the first line, all other lines should be changed to replace "pick" with the word "squash" instead.

Once this change has been made, save the file (:wq). This will cause another file to be opened in 'vi', containing the comments from all of the commits. Edit the file so that only a single comment remains that fully explains the changes that have been made.

Let me know if that doesn't make sense. :)

EricWittmann:
Note:  I think this change is good, but I wanted to make sure that you were also aware that the API Key generator is customizable.  You can provide your own custom impl of the IApiKeyGenerator interface via a plugin.

EricWittmann:
Great thanks!  As soon as Travis gives the OK I'll merge this in. :)

Thanks for the contribution.

ade90036:
I swashed the commits into one.

I'm aware of the ability to create our own key generator, however in our
scenario the key is generated from another system.

Regards

Andrea

EricWittmann:
Also note, I have created this JIRA for tracking purposes:

https://issues.jboss.org/browse/APIMAN-912

Once merged I will mark that jira as resolved.

ade90036:
Hi Eric,

this ticket: https://issues.jboss.org/browse/APIMAN-912 and your comments
got me thinking.

Can we not have the same apikey used on multiple contracts on the same
application?

We are trying to have the same apikey per application, which would simplify
authorization/authentication for service enpoint and SDK generations, given
the apikey is the same for multiple services in the app.

Regards

Andrea

EricWittmann:
The API Key really must be unique, because when a Request is received by the Gateway, it uses the API Key to look up the Contract. From the Contract we then know the ClientApp and Plan being used.  We currently allow a ClientApp to create more than one Contract to the same API.

In order to support an API Key per ClientApp and allow it to be re-used across multiple Contracts, we would need to re-architect the Gateway a little bit.

ade90036:
oh.. crap..

Is this a big change?

Is it something that i can manage?

EricWittmann:
It's a significant logical change - I'm not sure what the level of difficulty is until I poke around in the code a bit.  I'll discuss it with the rest of the team tomorrow...

ade90036:
Hi Eric,

Can you keep me informed on this development, discussions, plans, etc.

It is an important feature for us and this issue has potential to disrupt
our current design and stack.

Worst case scenario we can potentially fork the project but i would like to
be as close to the original project as possible.

Regards

Andrea

EricWittmann:
Yes - I'll try to report back here.  If you don't hear anything feel free to give a shout. :)

ade90036:
Thanks Eric..

ahilan-orlando:
Eric, I see a great value in simplifying the API key management. Andrea's feedback seems to align with our organization needs to effectively manage digital content to our partners across various channels.

I would like to see if this can be taken up as a new enhancement request.

Thank you !

EricWittmann:
Thanks for giving this thread a poke, @ahilan-orlando - as it turns out, we have decided to simplify the API Key management so that keys are generated for Client Apps rather than for each Contract.  This will have the effect of simplifying things (each Client App will be granted exactly one key).  However it will also mean that a Client App will only be allowed to create one Contract per API.  So even if an API is offered through multiple Plans, each Client App MUST choose one and only one Plan to use.

We think this is an acceptable compromise, and it's starting to sound like most of our community agrees.  :)

EricWittmann:
This change should happen "soon".  Most likely in the next two weeks or so.  It should be part of the next release (1.2.2.Final).  

Note that this will be a data model change, so it will require a change to the DB schema.  As part of this work, we will be creating an offline tool which will migrate exported apiman data from version 1.2.1.Final to 1.2.2.Final.  So the upgrade process will look like this:

1) Export all data from 1.2.1.Final
2) Shut down 1.2.1.Final
3) Run the offline data migration tool to upgrade the exported data (JSON) to the new format
4) Start up 1.2.2.Final
5) Import the migrated data into 1.2.2.Final

PR #335, opened 2016-01-21
use a route to get to the service definition url

KurtStam:


manager/ui/hawtio/plugins/api-manager/ts/catalog/api-catalog.ts
```diff
@@ -60,6 +60,9 @@ module Apiman {
                     if (api.endpointType == 'soap') {
                         api.ticon = 'fa-file-code-o';
                     }
+                    if (api.routeDefinitionUrl != null) {
+                        api.definitionUrl = api.routeDefinitionUrl;
+                    }
                 });
```
EricWittmann: So what's the point of returning both definition URLs if the routeDefinitionURL overrides the definitionURL if it exists?  E.g. when routeDefinitionURL then definitionURL isn't used (cannot be used because it is overwritten).

manager/api/beans/src/main/java/io/apiman/manager/api/beans/summary/AvailableApiBean.java
```diff
@@ -35,10 +35,12 @@
     private String id;
     private String icon;
     private String endpoint;
+    private String routeEndpoint;
```
EricWittmann: This got added to the bean but is never used in the UI.  What is it used for?

KurtStam: I think the UI should probably add this as a link (when unmanaged) so the user can get to the service. An example would be http://myservice.vagrant.f8. When the service is imported we'd advice to remove the route, and it should then link to the service endpoint on the gateway if possible.

manager/ui/hawtio/plugins/api-manager/ts/catalog/api-catalog.ts
KurtStam: No point in fabric8, but it might make sense on other platforms that don't support routes all together?

EricWittmann:
https://issues.jboss.org/browse/APIMAN-915

PR #336, opened 2016-01-21
Ignore Resources Policy: enhance to allow shading by VERB

wtrocki:
## What was done
- Added verb (http method) support for Ignore resources policy rule.
- Adjusted tests to cover new cases. 
- Reshaped UI to support new functionalities. 

## Gui changes

`IgnoredAligators`
http://gfycat.com/ThatBowedAlligatorgar

`AuthBorer`
http://gfycat.com/AliveKindlyBorer

## Technical comments

1. Policies gui components
I think we may improve a little bit the way we style policies. 
Currently we use nested style (which is slightly invalid) or direct style on elements. 
I think that we can have separate style templates for this (with gulp support). 
It might look better to clean all rules at the same time so I'm not pushing that here.

2. Separate ts files per policy.
It may be better to have separate ts files per policy (this may speed up gulp building)
Do we have any reasons to keep it in one file?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ gwt-unitCache
 *.iml
 .idea
 npm-debug.log
+.metadata
--- a/gateway/engine/policies/src/main/java/io/apiman/gateway/engine/policies/IgnoredResourcesPolicy.java
+++ b/gateway/engine/policies/src/main/java/io/apiman/gateway/engine/policies/IgnoredResourcesPolicy.java
@@ -15,23 +15,25 @@
  */
 package io.apiman.gateway.engine.policies;
 
+import io.apiman.gateway.engine.policies.config.IgnoredResource;
 import io.apiman.gateway.engine.beans.PolicyFailure;
 import io.apiman.gateway.engine.beans.PolicyFailureType;
 import io.apiman.gateway.engine.beans.ApiRequest;
 import io.apiman.gateway.engine.components.IPolicyFailureFactoryComponent;
+import io.apiman.gateway.engine.policies.PolicyFailureCodes;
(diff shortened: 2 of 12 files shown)


wtrocki:
Tests are passing locally. Investigating that at the moment.

EricWittmann:
The travis tests are failing due to a problem downloading one of the dependencies needed for the UI - it's not something you broke.

@msavy - do you have the link handy that describes how @wtrocki can workaround this?

wtrocki:
@EricWittmann Thanks for response. Actually this showed me that we have no unit test setup for UI. Is this something we are interested in?

msavy:
The only thing I can really suggest at the moment is that you enable TravisCI for your fork and test it there.

Sign up on travis-ci.org and then flip your `apiman` repo to be a candidate for testing. You may need to force re-push you branch to trigger the testing :).

EricWittmann:
@wtrocki - yes I think UI unit testing is interesting.  However, @kahboom is tasked with building a separate Developer Portal UI app, which will include unit tests and likely other common UI requirements that we missed when we implemented the manager UI.  Sadly those of us who implemented the manager UI don't have that much experience doing it. :)

So anyway - I've been waiting to see what @kahboom comes up with before retrofitting unit testing into the manager UI.

EricWittmann:
Sorry I haven't gotten to this yet.  I'll have a look on Monday and get it merged.

EricWittmann:
Hi @wtrocki  -  sorry for the delay, but I finally got around to merging your PR.  Thanks!

I made a few minor changes to the UI - elements weren't lining up very well in Firefox.  Hopefully should be all good now though.

PR #339, opened 2016-01-22
https://issues.jboss.org/browse/APIMAN-914 adding initialize to the I…

KurtStam:
…Storage interface

### DIFF
--- a/manager/api/core/src/main/java/io/apiman/manager/api/core/IStorage.java
+++ b/manager/api/core/src/main/java/io/apiman/manager/api/core/IStorage.java
@@ -53,6 +53,7 @@ public interface IStorage {
     public void beginTx() throws StorageException;
     public void commitTx() throws StorageException;
     public void rollbackTx();
+    public void initialize();
 
     /*
      * Various creation methods.  These are called by the REST layer to create stuff.
--- a/manager/api/es/src/main/java/io/apiman/manager/api/es/EsStorage.java
+++ b/manager/api/es/src/main/java/io/apiman/manager/api/es/EsStorage.java
@@ -151,6 +151,7 @@ public class EsStorage implements IStorage, IStorageQuery {
     /**
      * Called to initialize the storage.
      */
+    @Override
     public void initialize() {
         try {
             esClient.execute(new Health.Builder().build());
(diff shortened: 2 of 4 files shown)


EricWittmann:
If initialize() is being added to the IStorage interface, then it should be called more generically.  Currently initialize() only happens when using elasticsearch.  Instead we would need to change the following:

https://github.com/apiman/apiman/blob/master/manager/api/core/src/main/java/io/apiman/manager/api/core/config/ApiManagerConfig.java#L225-L227

And also everywhere we're currently calling initialize() only when it's ES - we should instead call it regardless of the impl:

ManagerApiMicroServiceCdiFactory
WarCdiFactory
TestCdiFactory

Thoughts?

KurtStam:
The issue is that in the CDI Factory case anyway, it gets called on
injection. If ES was not ready at that time there is no way to try again,
other then for the user to destroy the pod. It works fine when everything
is up and running.

So in the PR call to retry the initialization really happens outside of any
of the factories you mention. (It happens in the fabric8-microservice
instead during the bootstrap process).

I guess I fail to see why you need it be called 'everywhere'. Though I
guess it'd be more consistent.

--K

On Fri, Jan 22, 2016 at 3:39 PM, Eric Wittmann <notifications@github.com>
wrote:

> If initialize() is being added to the IStorage interface, then it should
> be called more generically. Currently initialize() only happens when using
> elasticsearch. Instead we would need to change the following:
>
>
> https://github.com/apiman/apiman/blob/master/manager/api/core/src/main/java/io/apiman/manager/api/core/config/ApiManagerConfig.java#L225-L227
>
> And also everywhere we're currently calling initialize() only when it's ES
> - we should instead call it regardless of the impl:
>
> ManagerApiMicroServiceCdiFactory
> WarCdiFactory
> TestCdiFactory
>
> Thoughts?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apiman/apiman/pull/339#issuecomment-174040096>.
>



-- 
Kurt T. Stam

twitter: @KurtStam
google+: kurt.stam@gmail.com

EricWittmann:
OK I can change it later I guess.

PR #340, opened 2016-01-24
Configurable connection options (timeouts, redirects) for service connector in 1.1.x

outofcoffee:
# Summary

* Adds the ability to control whether the gateway automatically follows redirects (301, 302 etc.) from the back-end service.
* Adds the ability to set the read timeout and connect timeout values for the service connection.

# Rationale

Prior to this change, the redirect behaviour of the gateway was the OkHttp default of 'follow redirects'. This can lead to some unexpected behaviour when back-end services return a redirect status code intended for the *client*, not the gateway.

The default setting (and the behaviour today) results in the gateway attempting to load the URL to which the client should be redirected, rather than passing through the 301/302 to the client. This change makes that behaviour configurable.

# Implementation

* A new `ConnectionOptions` class has been added that is fed by the engine configuration.
* The `HttpServiceConnection` is passed the connection options and applies them to the OkHttp client.
* The default values for timeouts and redirects have been set to ensure backwards compatibility for anybody relying on this behaviour today.

The new properties look like this:

```
# Connection settings for the gateway connector(s).

# Whether to automatically follow redirects (301, 302 etc.) from the
# back-end service.
apiman-gateway.connector-factory.connection.followRedirects=true

# Configure the timeout values for connections to the back-end service.
apiman-gateway.connector-factory.connection.readTimeout=15000
apiman-gateway.connector-factory.connection.connectTimeout=10000
```

## Considerations

There may be an argument for changing the default redirect-follow behaviour to 'false', to help avoid violating the principle of least surprise, but I leave this up to you @EricWittmann, @msavy.

### DIFF
--- a/common/config/src/main/java/io/apiman/common/config/options/AbstractOptions.java
+++ b/common/config/src/main/java/io/apiman/common/config/options/AbstractOptions.java
@@ -19,6 +19,7 @@ import java.util.Map;
 
 import org.apache.commons.lang.BooleanUtils;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.math.NumberUtils;
 
 /**
  * Base class for all options.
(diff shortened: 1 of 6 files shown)


outofcoffee:
@EricWittmann, @msavy this is a backport of https://github.com/apiman/apiman/pull/341 to the `1.1.x` branch, as well as some of the other connector properties stuff already in `master`. If you're not maintaining the `1.1.x` branch anymore, then you probably don't need this PR :)

EricWittmann:
We don't have any plans to release another 1.1.x version - but it sure doesn't hurt to have this there in case we do.

PR #341, opened 2016-01-25
Configurable redirect follow behaviour for gateway

outofcoffee:
# Summary

Adds the ability to control whether the gateway automatically follows redirects (301, 302 etc.) from the back-end service.

# Rationale

Prior to this change, the redirect behaviour of the gateway was the OkHttp default of 'follow redirects'. This can lead to some unexpected behaviour when back-end services return a redirect status code intended for the *client*, not the gateway.

The default setting (and the behaviour today) results in the gateway attempting to load the URL to which the client should be redirected, rather than passing through the 301/302 to the client. This change makes that behaviour configurable.

# Implementation

* A new property has been added to `HttpConnectorOptions`.
* The `HttpConnectorFactory` applies the redirect property to the OkHttpClient.
* The default value for redirects has been set to ensure backwards compatibility for anybody relying on this behaviour today.

The new property looks like this:

```
# Connector factory options

# Whether to automatically follow redirects (301, 302 etc.) from the
# back-end service.
apiman-gateway.connector-factory.http.followRedirects=true
```

## Considerations

There may be an argument for changing the default redirect-follow behaviour to 'false', to help avoid violating the principle of least surprise, but I leave this up to you @EricWittmann, @msavy.

### DIFF
--- a/common/config/src/main/java/io/apiman/common/config/options/HttpConnectorOptions.java
+++ b/common/config/src/main/java/io/apiman/common/config/options/HttpConnectorOptions.java
@@ -28,10 +28,12 @@ public class HttpConnectorOptions extends AbstractOptions {
     private static final int DEFAULT_READ_TIMEOUT = 30;
     private static final int DEFAULT_WRITE_TIMEOUT = 30;
     private static final int DEFAULT_CONNECT_TIMEOUT = 10;
-    
+    private static final boolean DEFAULT_FOLLOW_REDIRECTS = true;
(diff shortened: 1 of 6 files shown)


EricWittmann:
I agree that redirect-follow behavior should default to 'false'. I doubt anyone is relying on the current default behavior.  I'll make the change after merging. :)

Thanks!

outofcoffee:
Cheers, Eric!

PR #346, opened 2016-01-27
Handle LDAP issue when error when certain errors when using ServiceAccount can be blackholed

msavy:


msavy:
cc @jcechace @xjaros1

EricWittmann:
Do we need to add a unit test for this?

msavy:
Yeah, and I think a bit of control flow needs changing - it's a bit awkward to the interface, but it seems in this case it'll cause a 500 error which is kinda uninformative. Will reopen when done.

jcechace:
Thanks for the mention, the fix should be verified tonight (when the ci job gets triggered)

msavy:
No Problem! I’m making some improvements and will reopen the commit later.

msavy:
OK, have revised this.

msavy:
@EricWittmann - OK should be good now

PR #347, opened 2016-01-29
Changes for APIMAN-942 - Ability to delete an unpublished service.

ctsantiago:
This is a proposed change described in:
https://issues.jboss.org/browse/APIMAN-942

### DIFF
--- a/manager/api/core/src/main/java/io/apiman/manager/api/core/IStorage.java
+++ b/manager/api/core/src/main/java/io/apiman/manager/api/core/IStorage.java
@@ -33,7 +33,10 @@ import io.apiman.manager.api.beans.plugins.PluginBean;
 import io.apiman.manager.api.beans.policies.PolicyBean;
 import io.apiman.manager.api.beans.policies.PolicyDefinitionBean;
 import io.apiman.manager.api.beans.policies.PolicyType;
+import io.apiman.manager.api.beans.summary.ApiVersionSummaryBean;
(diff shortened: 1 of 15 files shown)


EricWittmann:
This is a great start, thanks!

*Note*: I am going to make the following changes to this PR:
* The endpoint to DELETE an API:
```
    @DELETE
    @Path("{organizationId}/apis/{apiId}")
    public void deleteApi(@PathParam("organizationId") String organizationId,
            @PathParam("apiId") String apiId)
            throws ApiNotFoundException, NotAuthorizedException, InvalidApiStatusException;
```
* The user will need to have the "apiAdmin" privilege to delete, rather than full Admin privs

EricWittmann:
OK @ctsantiago  - merged and modified.  Please have a look and let me know if the solution still works for you! :)

ctsantiago:
Thanks Eric.  I learned a lot from your changes for next time.  This works for us.

EricWittmann:
OK great!  I think there is still work to be done on the Delete API feature - but it should be a good start.  For example, I'd like it to work provided there are *no* published APIs.  But that means when an API is retired we need to break all its contracts.  Or else we need to break the contracts when deleting it (that's probably the way to go).

But for now this gets us some common use cases (e.g. I created an API and never published it, but now I can't delete it!).  :)

We also still need to work on deleting Plans, Orgs, and Clients.  Again, I think this PR is a good start to all that work.

PR #361, opened 2016-02-08
APIMAN-953: make apiman able to process GET parameters with multiple …

bramvonk:
…values

APIMAN-953: make apiman able to process GET parameters with multiple
values.
Backwards-compatible for *reading* get parameters the old way
(returns the last value, just as with the older implementation), but
trying to delete/add/alter the old map throws an
UnsupportedOperationException.
Also fixes a problem with processing
parameter encoding. Added tests for that.


EricWittmann:
Closing in favor of:  https://github.com/apiman/apiman/pull/363

@bramvonk can you re-issue a separate PR with just your query param urldecode bug fix?

PR #365, opened 2016-02-08
QueryMap #toString and #toQueryString that builds www-form-urlencoded query string

msavy:
@EricWittmann as requested... :smile:

### DIFF
--- a/gateway/engine/beans/src/main/java/io/apiman/gateway/engine/beans/util/QueryMap.java
+++ b/gateway/engine/beans/src/main/java/io/apiman/gateway/engine/beans/util/QueryMap.java
@@ -16,6 +16,11 @@
 
 package io.apiman.gateway.engine.beans.util;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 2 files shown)


gateway/engine/beans/src/main/java/io/apiman/gateway/engine/beans/util/QueryMap.java
```diff
@@ -33,4 +38,26 @@ public QueryMap() {}
...
+
+    @Override
+    public String toString() {
+        return toQueryString();
+    }
+
+    @SuppressWarnings("nls")
+    public String toQueryString() {
+        return getEntries().stream()
+                .map(pair -> URLEnc(pair.getKey()) + "=" + URLEnc(pair.getValue()))
```
msavy: @EricWittmann @kahboom I spent a while scouring the internet, and it seems that `www-form-urlencoding` is valid for the query part of the URL (which is what Java's confusingly named `URLEncoder` does. 

This is slightly different from *actual* URLEncoding which uses `%20` instead of `+` for `[:space:]`.

My only concern is that some people might be bothered if we swap between the different encoding strategies. IMO this shouldn't be an issue, but we should keep it in mind.

PR #372, opened 2016-02-09
Fix remaining transitive issues and enable enforcement of transitive errors

msavy:


msavy:
Seems I had some caching that means I missed some things :S

PR #375, opened 2016-02-10
Fix remaining transitive issues and enable enforcement of transitive dep flaws

msavy:
cc @rnc 

BTW I also removed some unused deps. We seem to have a fair few, but it's a pain to figure out because the command to find unused deps often conflicts with the command that figures out transitive deps. It's a decent starting point, though.

```
mvn dependency:analyze -DignoreNonCompile=true 
```

Also, I found a neat plugin to sort pom files in a sensible way automatically. Here's the command I tried out that imitates our usage. I've not applied it broadly, just FYI.

```
mvn com.github.ekryd.sortpom:sortpom-maven-plugin:sort -Dsort.keepBlankLines -Dsort.predefinedSortOrder=custom_1 -Dsort.sortProperties=true -Dsort.sortDependencies=scope,groupId,artifactId
```

### DIFF
--- a/gateway/engine/es/pom.xml
+++ b/gateway/engine/es/pom.xml
@@ -97,8 +97,24 @@
       <artifactId>commons-lang</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>commons-io</groupId>
+      <artifactId>commons-io</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>commons-codec</groupId>
+      <artifactId>commons-codec</artifactId>
+    </dependency>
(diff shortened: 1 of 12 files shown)


msavy:
BTW @EricWittmann @kahboom - if you ever have the need to find these transitive deps and add them, it can be a huge pain in the butt figuring out which GAV you need to import to get the classes you need.

Finally, last night I discovered that maven search actually lets you search by class, and reverse maps which packages contain that class.

go to search.maven.org and type `fc:"<path to class>"`

e.g.

```
fc:"io.apiman.manager.api.exportimport.manager.ExportImportManager"

GroupId
io.apiman	

ArtifactId
apiman-manager-api-export-import-api	

Version
1.2.1.Final

Updated
21-Jan-2016
<etc ...>
```

EricWittmann:
That's super cool, thanks for the tip.

PR #376, opened 2016-02-10
APIMAN-970 parameter decoding

bramvonk:
Fix for handling parameters with & and = encoded in its values or keys.


EricWittmann:
Thanks!

PR #377, opened 2016-02-10
Update README.md

msavy:
@EricWittmann @kahboom Do these changes seem reasonable? In particular the new bit about squashing commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,11 @@ The general idea is to keep your 'master' branch in-sync with the 'upstream/mast
 
 ## Building apiman
 
-####Note: If you're using JDK 1.7 or earlier, please allocate additional memory within your user profile to allow Maven to do its work:
-  
-	Linux/OS X:
-		export MAVEN_OPTS="-Xmx1024m -XX:MaxPermSize=256M"
-	Windows:
-		set MAVEN_OPTS="-Xmx1024m -XX:MaxPermSize=256M"
+### Requirements
+- Maven 3.x
+- Java 8+
 
-We use Maven 3.x to build our software. The following command compiles all the code, installs the JARs into 
-your local Maven repository, and runs all of the unit tests:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,16 +40,16 @@ your local Maven repository, and runs all of the unit tests:
 The project can be built and deployed on a variety of runtime platforms, but if you want to see it in 
 action as quickly as possible try this:
 
-    $ mvn clean install -Pinstall-all-wildfly8
-    $ cd tools/server-all/target/wildfly-8.2.0.Final/
+    $ mvn clean install -Pinstall-all-wildfly9
```
EricWittmann: Should actually be wildfly10 now I think...

EricWittmann:
Yes, I like the part about squashing commits.  I'll try to start pushing back on that more.

README.md
msavy: I wasn't too sure about this, simply because of the relative URLs bug with KC. Or is that also in WF9?

msavy: @EricWittmann Happy to do either. Especially as we'll be upgrading to a version of KC with fixed relative URL handling.

EricWittmann: Yes, it should be 10.  The bug is with KC and has nothing to do with WF.  To fix the bug we need to upgrade to KC 1.9.

msavy: Yeah, I just thought we used a different version of KC with the WF9 vs WF10 and the version we were using with WF10 was the b0rked one. Anyway, I realise that's incorrect now :o).

Updated!

EricWittmann: The diff still shows WF 9 everywhere.  :)

msavy: Bash F5 a few times :o)

msavy: Well, it's updated in my repo (https://github.com/msavy/apiman/blob/readme-java8/README.md) but for some reason the change isn't reflecting here. Maybe I'll open a new PR.

PR #381, opened 2016-02-11
Apiman OSGI UI bundle

cmoulliard:
- Refactor the project to create the Apiman OSGI UI bundle
- Add the option "attachclasses" option to the maven war plugin of Apiman War ui in order to generate the jar containing the code of the project
- Expose IUIConfig as OSGI Service
- Include an Activator to register the different servlets
- Add missing packages


manager/ui/war/apiman/config.js-SAMPLE
```diff
@@ -12,7 +12,7 @@ var APIMAN_CONFIG_DATA = {
     "metrics" : true
   },
   "api" : {
-    "endpoint" : "http://localhost:8080/apiman",
+    "endpoint" : "http://localhost:8181/apimanui/apiman",
```
EricWittmann: Please roll this change back - the sample config should be 8080.  Also not sure why it would ever be /apimanui/apiman?

manager/ui/war/src/main/java/io/apiman/manager/ui/server/impl/UIConfig.java
```diff
@@ -13,9 +13,10 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.apiman.manager.ui.server;
+package io.apiman.manager.ui.server.impl;
```
EricWittmann: Why this change?

cmoulliard: To avoid that the classloader of apiman-common-util imports the impl class located here previsouly "io.apiman.manager.ui.server" as we will use another class for the osgi bundle of the apiman manager ui

manager/ui/war/apiman/config.js-SAMPLE
cmoulliard: ok. will do that

PR #385, opened 2016-02-13
APIMAN-688 - TimeRestrictedAccessPolicy

wtrocki:
#  TimeRestrictedAccessPolicy

Initial implementation of the new policy that restrict access to the API basing on specified time.
Name and behavior open for suggestions.

##  How it looks
http://www.gfycat.com/ContentFormalDodobird

##  How it works.
Accepts requests that only in specified time range and days of the week. 
Multiple rules that matching the same pattern are joined together at the execution time.
User can specify 2 different ranges and times for the same pattern. 
Two weeks cycles are supported  - from FRI to MON etc.

JSON data is stored with very extendable format (cron style)
<pre>
'timeStart'  (DATE)
'timeEnd' :  (DATE)
'dayStart' :  1-7
'dayEnd' 1-7
'pathPattern' (regexp)
</pre>

We can easily extend it to week.month, year if we need it. For exact ranges we can use timeStart,timeEnd - but this is not currently supported in gui.


EricWittmann:
This looks really great, thanks @wtrocki .  One bit of feedback about policy unit testing.  What you have here is fine but uses an older approach (mockito).  Nothing wrong with that, but we created a really nice junit framework for testing policies.  Here are some examples of it in action:

* https://github.com/apiman/apiman/blob/master/gateway/engine/policies/src/test/java/io/apiman/gateway/engine/policies/BasicAuthenticationPolicyTest.java
* https://github.com/apiman/apiman/blob/master/gateway/engine/policies/src/test/java/io/apiman/gateway/engine/policies/TransferQuotaPolicyTest.java
* https://github.com/apiman/apiman/blob/master/gateway/engine/policies/src/test/java/io/apiman/gateway/engine/policies/CachingPolicyTest.java

And here is the documentation for how to unit test a policy:

http://www.apiman.io/latest/developer-guide.html#_unit_testing_a_plugin_policy

No reason to change what you have here unless you want to.  But thought I would mention it.

PR #388, opened 2016-02-15
Upgrade to json-editor 0.7.23 to fix spinner issue in APIMAN-544

msavy:
@kahboom Please verify this doesn't break any of the stuff you were working on lately! 

Fixes https://issues.jboss.org/browse/APIMAN-544

### DIFF
--- a/manager/ui/war/npm-shrinkwrap.json
+++ b/manager/ui/war/npm-shrinkwrap.json
@@ -2686,9 +2686,9 @@
       }
     },
     "json-editor": {
-      "version": "0.7.21",
-      "from": "json-editor@0.7.21",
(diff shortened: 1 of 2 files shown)


kahboom:
@msavy - Thanks, I tested and this doesn't break the update here: https://github.com/jdorn/json-editor/issues/547, but there is a slight difference in the UI (I don't remember it being like this before, but please correct me if I'm wrong):

So, if you add a value and delete it, and then go to add a new value again, it will highlight it in red like if the user had done something wrong. This is very minor though and I don't think we should worry about it.

screen shot 2016-02-15 at 11 36 41 am

msavy:
Having an empty field is invalid - albeit the error message isn’t helpful.

Hopefully once we have full custom UIs for plugins we can do away with this stuff.

PR #389, opened 2016-02-15
re #968 Swap method and path on table

kahboom:
Swapped the "method" and "path" labels on the table header of the Ignored Resources Policy Configuration page.

JIRA: https://issues.jboss.org/browse/APIMAN-968

screen shot 2016-02-15 at 2 39 05 pm

cc @EricWittmann

### DIFF
--- a/manager/ui/war/plugins/api-manager/html/policyForms/ignored-resources.include
+++ b/manager/ui/war/plugins/api-manager/html/policyForms/ignored-resources.include
@@ -37,8 +37,8 @@
   <table class="table table-striped table-bordered">
     <thead>
       <tr>
-        <th width="30%" apiman-i18n-key="verb">Method</th>
-        <th width="70%" apiman-i18n-key="path">Path</th>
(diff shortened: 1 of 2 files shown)


manager/ui/war/plugins/api-manager/html/policyForms/ignored-resources.include
```diff
@@ -37,8 +37,8 @@
   <table class="table table-striped table-bordered">
     <thead>
       <tr>
-        <th width="30%" apiman-i18n-key="verb">Method</th>
-        <th width="70%" apiman-i18n-key="path">Path</th>
+        <th width="30%" apiman-i18n-key="path">Path</th>
+        <th width="70%" apiman-i18n-key="verb">Method</th>
```
EricWittmann: Could you also change the i18n key to be "method"?

kahboom: Done! Also forgot about the favicon change for the JSON schema viewer last week, so added that in here as well.

PR #390, opened 2016-02-16
Override hashcode whenever overriding equals() and "equals(Object obj)" should test argument type

AymanDF:
This pull request is focused on resolving occurrences of FindBugs rule HE_EQUALS_USE_HASHCODE - “ Class defines equals() and uses Object.hashCode() ”. You can find more information about the issue here: http://findbugs.sourceforge.net/bugDescriptions.html#HE_EQUALS_USE_HASHCODE
And
Sonar rule squid:S2097 - “ "equals(Object obj)" should test argument type ”. You can find more information about the issue here:
https://dev.eclipse.org/sonar/rules/show/squid:S2097
Please let me know if you have any questions.
Ayman Abdelghany.


EricWittmann:
Good change, thanks.  The irony of it, however, is that this code violates the "unnecessary parentheses" rule from a previous PR!  ;)

It's OK though since I think we're likely to reject most of the unnecessary parens PR.

PR #405, opened 2016-02-19
Batch of quality fixes

msavy:
The only ones of real note are some potential null accesses,
and dangling resources.

Used SonarQube...


manager/api/export-import/src/main/java/io/apiman/manager/api/exportimport/json/JsonImportReader.java
```diff
@@ -101,52 +101,22 @@ public void read() throws Exception {
 
                 switch(fieldName) {
                 case Metadata:
-                    processEntity(MetadataBean.class, new EntityHandler<MetadataBean>() {
-                        @Override
-                        public void handleEntity(MetadataBean metadata) throws Exception {
-                            dispatcher.metadata(metadata);
-                        }
-                    });
+                    processEntity(MetadataBean.class, dispatcher::metadata);
```
EricWittmann: Now here is a big syntactic win!

msavy: +1 - much cleaner!

PR #406, opened 2016-02-19
(Bug) Fix resource leak. StringBuilderWriter never closed.

msavy:


EricWittmann:
A stringbuilderwriter doesn't need to be closed but this change is fine anyway. :-)

msavy:
Oh, you're right.

PR #407, opened 2016-02-19
(Bug) Fix multiple resource leaks in ES Marshalling code - *please review*

msavy:
Tests pass OK, and from what I can ascertain this doesn't change any intended behaviour, but as I'm not the original author of this I just want someone to review it to check.

The issue is that the XContentBuilder is never closed, so its underlying builder is, in turn, never closed.

@EricWittmann Please check this out.

### DIFF
--- a/manager/api/es/src/main/java/io/apiman/manager/api/es/EsMarshalling.java
+++ b/manager/api/es/src/main/java/io/apiman/manager/api/es/EsMarshalling.java
@@ -94,9 +94,9 @@ public class EsMarshalling {
      * @throws StorageException when a storage problem occurs while storing a bean
      */
     public static XContentBuilder marshall(PoliciesBean bean) throws StorageException {
-        try {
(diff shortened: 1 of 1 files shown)


msavy:
This is potentially the source of a whole bunch of leaks due to the number of code-paths using this stuff in addition to sheer volume.

msavy:
Looking at the code of `XContentBuilder` it seems that most of the methods will cause `close` on your behalf. So perhaps it isn't needed?

EricWittmann:
Yeah I assume this thing is just backed by a string builder, but it's safe enough to close it I imagine.

msavy:
Digging into the code - it seems to be backed by the Jackson JSON generator, so it does need to be closed explicitly. However, it does seems to try to close itself when you call various terminal methods (e.g. toString). 

The change should catch any cases when an exception might cause a terminal operation to be skipped.

So should be useful (hopefully!).

PR #411, opened 2016-02-21
(Bug) Broken string equality comparison in TransferQuotaPolicy

msavy:
`==` vs `equals` strikes again.

Really, I wish Java was sensible on this like most other languages. 99% of the time people don't want to compare object references!


EricWittmann:
Sorry, this is as-designed.  The == comparison is appropriate in this case.

msavy:
Ah, I see what you did in `#bucketId`. Apologies for the noise!

PR #412, opened 2016-02-21
(Chore) Refactor common code in api/rest/impl

msavy:
Some fair chunks of code were repeated, so extracted into AbstractResourceImpl


gateway/api/rest-impl/src/main/java/io/apiman/gateway/api/rest/impl/IPlatformAccessor.java
```diff
@@ -31,5 +31,5 @@
      * Gets the platform.
      * @return the engine
      */
-    public IPlatform getPlatform();
+    IPlatform getPlatform();
```
EricWittmann: I know public is redundant, but I like it.  :)  Unless there's a compelling reason to remove it I'd prefer to leave it alone.

msavy: ok

PR #415, opened 2016-02-23
re #992 Switch from TSD to Typings

kahboom:
Changes:
- Remove `typings` dir and `tsd.json` file locally.
- Uninstall TSD, install Typings.
- Upgrade TSD JSON file to Typings JSON file.
- Remove TSD JSON file.
- Remove `gulp-tsd` package, Gulp 'setup' task, instead ask NPM to use TS binary to install TS definitions.
- Rewrite `npm-shrinkwrap.json` file.
- Update Gulp to `3.9.1` latest patch.

JIRA: https://issues.jboss.org/browse/APIMAN-992

cc @EricWittmann @msavy  - Tested on OS X and everything looked good, but please let me know if you run into anything strange. Locally, you should be safe to delete the `typings` dir and `tsd.json` file. You may need to run `rm -rf node_modules & npm install`.


msavy:
I thought we were committing all of the TSD dips into the repo? This was one of the prod requirements. 

Can we achieve this?

E.g. install TSD with typings, then commit installed definitions to the repo; presumably we could then point our build to aforementioned directory (or use typings and refer it to that directory?)

kahboom:
@msavy - Why were we looking to check in TS definitions again? Was this in order to make 100% offline installations possible (ie. to avoid the GH rate-limiting issue)? And np I'll commit the `typings` dir and remove it from `.gitignore`. We are already committing the `typings-custom` dir.

msavy:
It’s related to prod and their current lack of tooling to handle Typings.

EricWittmann:
@kahboom - yes, the reason to check in the TSD files is so that prod builds don't have to download them.

Let me know when this is ready to be merged based on marc's feedback.

kahboom:
@EricWittmann @msavy - Done! Let me know if I missed anything.

PR #424, opened 2016-02-29
Profile out further modules

rnc:
This is a continuation of SHA 53c1310ab81337e8658f37ac7bed180ea9661d0a.

Building as a non-snapshot version reveals that these modules are dependent on the distros.

### DIFF
--- a/gateway/pom.xml
+++ b/gateway/pom.xml
@@ -14,6 +14,19 @@
     <module>api</module>
     <module>engine</module>
     <module>platforms</module>
-    <module>test</module>
   </modules>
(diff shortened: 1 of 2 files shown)


EricWittmann:
Ah sorry about that.  However, looking at it again I think a different solution is required.  I'm going to reject this PR but will fix this immediately in a slightly different way.

EricWittmann:
@rnc see here:  https://github.com/apiman/apiman/commit/97ebf192cffc997dc98d0865b52ccfef825ed19c

PR #427, opened 2016-03-02
re #996 Delete APIs in UI

kahboom:
Changes:
- Add modal template and button to delete API (let me know if you don't like the cog dropdown option. I personally like it but we can go the traditional button route, I just didn't want to clutter the UI or "encourage" the user to delete).
- Add request to API to delete API
- Add comparison check for user input vs name of API after forcing both as lowercase, for user convenience
- Quick fix for How to Invoke API modal close button not working (JIRA: https://issues.jboss.org/browse/APIMAN-1001)
- Minor styling (mostly inline for now, will be improved later)

Also, in the future we need to implement a check that determines whether to show/hide the delete button at all in the UI. For now, the user will know because they will get an error when they try to delete an "undeletable" API.

Screenshots:

screen shot 2016-03-02 at 11 52 34 am

screen shot 2016-03-02 at 12 02 46 pm

screen shot 2016-03-02 at 12 02 55 pm

JIRA: https://issues.jboss.org/browse/APIMAN-996

cc @EricWittmann @msavy

### DIFF
--- a/manager/ui/war/plugins/api-manager/ts/api/api.ts
+++ b/manager/ui/war/plugins/api-manager/ts/api/api.ts
@@ -51,8 +51,8 @@ module Apiman {
         }]);
 
     export var ApiEntityController = _module.controller("Apiman.ApiEntityController",
-        ['$rootScope', '$q', '$location', '$scope', 'ActionSvcs', 'Logger', 'Dialogs', 'PageLifecycle', '$routeParams', 'OrgSvcs', 'EntityStatusSvc', 'Configuration',
-        ($rootScope, $q, $location, $scope, ActionSvcs, Logger, Dialogs, PageLifecycle, $routeParams, OrgSvcs, EntityStatusSvc, Configuration) => {
+        ['$rootScope', '$q', '$location', '$scope', '$uibModal', 'ActionSvcs', 'Logger', 'Dialogs', 'PageLifecycle', '$routeParams', 'OrgSvcs', 'EntityStatusSvc', 'Configuration',
+        ($rootScope, $q, $location, $scope, $uibModal, ActionSvcs, Logger, Dialogs, PageLifecycle, $routeParams, OrgSvcs, EntityStatusSvc, Configuration) => {
             var params = $routeParams;
             $scope.params = params;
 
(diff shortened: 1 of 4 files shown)


manager/ui/war/plugins/api-manager/ts/api/api.ts
```diff
@@ -88,6 +88,33 @@ module Apiman {
...
+                        params: function() {
+                            return $scope.params;
+                        }
+                    }
+                });
+
+                modalInstance.result.then(function (selectedItem) {
+                    $scope.selected = selectedItem;
+                }, function () {
+                    $log.info('Modal dismissed at: ' + new Date());
```
EricWittmann: Don't use $log - instead use Logger, which is already being injected.

manager/ui/war/plugins/api-manager/ts/api/api.ts
```diff
@@ -196,4 +223,53 @@ module Apiman {
...
+        $scope.okayToDelete = false;
+
+        $scope.typed = function () {
+            // For user convenience, compare lower case values so that check is not case-sensitive
+            $scope.okayToDelete = ($scope.confirmApiName.toLowerCase() === api.name.toLowerCase());
+        };
+
+        // Yes, delete the API
+        $scope.yes = function () {
+            console.log('Preparing for delete.');
```
EricWittmann: Please use Logger instead.  And maybe use debug() instead of log()?

EricWittmann:
This looks really great!  Nice job.

Does it actually work?! :)

manager/ui/war/plugins/api-manager/ts/api/api.ts
kahboom: +1 will update this in the morning. I'm off but will be around and this should be a quick change.

kahboom:
@EricWittmann - Well, I get a "method not allowed" error when trying to delete, if that's what you mean (see screenshot below). I couldn't test this with mock requests because I don't have unit tests set up, and the backend isn't ready yet. Or should this request have succeeded? I tested manually with a published API and an API that has never been published before (with only one version).

screen shot 2016-03-02 at 8 44 12 pm

EricWittmann:
Ah yes I see - you're not sending the correct REST request.  You're sending a request here:

```
/apiman/actions
```
https://github.com/apiman/apiman/blob/master/manager/api/rest/src/main/java/io/apiman/manager/api/rest/contract/IActionResource.java#L34

Instead, you want to send the DELETE request here:

```
/apiman/organizations/{orgId}/apis/{apiId}
```
https://github.com/apiman/apiman/blob/master/manager/api/rest/src/main/java/io/apiman/manager/api/rest/contract/IOrganizationResource.java#L825-L829

I believe it *should* work.

kahboom:
Wow lol, thanks.

EricWittmann:
Sorry - I was waiting for this to be updated with the proper REST api call.  You did update it but I didn't get a notification of that change.  So now it's merged!

PR #429, opened 2016-03-03
making the ES index names configurable https://issues.jboss.org/brows…

KurtStam:
…e/APIMAN-1010


EricWittmann:
Thanks for the PR - unfortunately I was already working on this and there's a bit of complexity that your PR misses. Specifically, we need to have both the index name *and* the default index name when we create the Jest client, because the latter is used when creating the index if it does not yet exist.  Here is the first part of the solution:

https://github.com/apiman/apiman/commit/b1d609bd9592023f44d00b11c82735940fda84ef

The rest of the solution, which is to actually allow the index names to be specified via configuration, I have not yet checked in.  :(  But it will be done today.

PR #435, opened 2016-03-17
Normalises HTTP Basic scheme name to match that of RFC1945

outofcoffee:
This PR normalises the HTTP Basic scheme name to match that of RFC1945.

See: http://tools.ietf.org/html/rfc1945#section-11.1

This fixes an issue with strict back-end services rejecting 'BASIC' (the current value) instead of 'Basic'.


EricWittmann:
Thanks!

PR #436, opened 2016-03-18
re #1025 - API Catalog: Ability to search for and filter with multiple tags

kahboom:
Changes:
- Add tag filter that supports multiple tags. Gather tags from available APIs. Sort tags alphabetically.
- Ability to clear all tag selections or individual tags.
- Support for all filters in combinations.
- Add UI-select tagging ability to search for any tags.
- CSS to match neighboring API type dropdown.

Tested on Chrome, Safari & FF running OS X El Capitan.

--

### Screenshots

**Default view:**
screen shot 2016-03-18 at 12 46 36 pm

**Focus opens up dropdown (see note about this below):**
screen shot 2016-03-18 at 12 46 51 pm

**Filter with one tag selected:**
screen shot 2016-03-18 at 12 47 06 pm

**Filter with multiple tags selected:**
screen shot 2016-03-18 at 12 47 19 pm

**Support for searching for tags:**
screen shot 2016-03-18 at 1 16 57 pm

**With all three filters being used:**
screen shot 2016-03-18 at 1 19 20 pm

--

JIRA: https://issues.jboss.org/browse/APIMAN-1025

cc @EricWittmann - Apologies for the delay. I know you didn't want a dropdown but it is included by default, so I wanted to show you how it looks with it. If you'd still prefer not to have it let me know and I'll see if I can hide it with CSS! :)


EricWittmann:
I actually *did* want the drop-down, but only if it wasn't a lot of work to implement.  So I'm perfectly happy with that.

Feedback/questions:

1) Sorry to be pedantic - but from the screen shot(s) it looks like the tag filter box is one (1) pixel taller than the other filters (same with the Clear button)
2) Does the impl support clicking on a tag on one of the cards to filter by that tag?

kahboom:
@EricWittmann - Great! Glad to hear that; I prefer it with the drop down as well. I fixed the first issue on my side, but then I noticed a difference in font size and padding between the two text boxes is driving me insane so I'm trying to fix it.

As for your second question-- yep! It supports clicking a tag on one of the cards to filter by that tag. The only thing I don't know, and will test now, is whether it will *remove* that tag from the filter if you click it twice. Not sure if that's even something we'd want or is worth implementing if it doesn't.

kahboom:
@EricWittmann - See screenshot of new changes:

screen shot 2016-03-18 at 6 27 55 pm


- Adjusted height of input field.
- Removed a box shadow from inside the input field to match the other input field.
- Fixed a CSS issue where the input field would get a blue border on hover or focus; changed it to match the other input field.
- CSS, match font size and padding of other input field.

And just to confirm, clicking on the tag a second time from the card does *not* remove it from the filter, which is the standard and makes sense. Let me know if you'd prefer otherwise though.

PR #446, opened 2016-04-06
[APIMAN-883] [APIMAN-890] [APIMAN-942] Delete feature (Org, API and ClientApp)

msavy:
Just been doing tonnes of tidying up (and squashing away a huge number of failed intermediary attempts!).

Please take a look @apiman/developers !

### DIFF
--- a/manager/api/jpa/src/main/java/io/apiman/manager/api/jpa/AbstractJpaStorage.java
+++ b/manager/api/jpa/src/main/java/io/apiman/manager/api/jpa/AbstractJpaStorage.java
@@ -95,7 +95,6 @@ public abstract class AbstractJpaStorage {
             throw new StorageException(e);
         } catch (RollbackException e) {
             logger.error(e.getMessage(), e);
-            e.printStackTrace(System.err);
             throw new StorageException(e);
         } catch (Throwable t) {
             logger.error(t.getMessage(), t);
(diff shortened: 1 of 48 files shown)


msavy:
Oh darn, I just realised I forgot to check for permissions. I'll definitely need to do that.

What are we thinking? Administrator?

msavy:
@KurtStam This may interest you. The interface is (roughly):

DELETE organizations/{organizationId}
DELETE organizations/{organizationId}/clients/{clientId}
DELETE organizations/{organizationId}/apis/{apiId}

Note that in order for a delete to be allowed a given entity must not be active. For instance, you can't delete an API if there are any published versions (retire any that are).

manager/api/jpa/src/main/java/io/apiman/manager/api/jpa/AbstractJpaStorage.java
```diff
@@ -95,6 +95,7 @@ protected void commitTx() throws StorageException {
             throw new StorageException(e);
         } catch (RollbackException e) {
             logger.error(e.getMessage(), e);
+            e.printStackTrace(System.err);
```
msavy: Whoops. I'll remove these.

msavy:
:ship:

PR #452, opened 2016-04-13
(Bug) Add workaround for MySQL shortfalls by using Native SQL.

msavy:
Unfortunately it doesn't seem possible to do these queries in any reasonably efficient way with JQPL that also works with MySQL's limitations on subqueries.

JIRA: APIMAN-1109


msavy:
Well, this is fun. MySQL likes the queries, but now H2 doesn't.

PR #455, opened 2016-04-20
[APIMAN-586] [APIMAN-946] (Feature) Make logging available to policies 

msavy:
Make logging available to policies.

Refactored existing logging support in the manager to make
it available to the gateway.

Some default implementations are available, but specific
implementations should likely be provided for each platform.

JIRA: APIMAN-586,  APIMAN-946


gateway/test/src/test/java/io/apiman/gateway/test/policies/LoggingPolicy.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * @see io.apiman.gateway.engine.policy.IPolicy#apply(io.apiman.gateway.engine.beans.ApiRequest, io.apiman.gateway.engine.policy.IPolicyContext, java.lang.Object, io.apiman.gateway.engine.policy.IPolicyChain)
+     */
+    @Override
+    @SuppressWarnings("nls")
+    public void apply(final ApiRequest request, final IPolicyContext context, final Object config,
+            final IPolicyChain<ApiRequest> chain) {
+        // This is just for testing, don't do this at home, kids.
+        TestLogger logger = (TestLogger) context.getLogger(getClass());
+        logger.setHeaders(request.getHeaders());
```
msavy: TestLogger writes to the headers instead of stdout (or whatever), so it can be tested against using resttests

msavy:
I'll get rid of 2aa5cba when we get feedback on how to do this best for WF/EAP. In the meanwhile I reckon looking at a simple Log4j2 logger makes sense IMO (essentially the same as the logger as I did for the manager). For instance, tomcat, jetty, etc.

Would need to ensure to use the async logger, which is mainly config but also requires an additional dep (https://logging.apache.org/log4j/2.x/manual/async.html)

EricWittmann:
+1

PR #459, opened 2016-04-23
Adds Tomcat logging provider.

outofcoffee:
# Background

At the moment, the gateway logs in Tomcat don't work, as there's no SLF4J `StaticLoggerBinder` on the classpath. Tomcat uses JUL as its default logging provider.

# This PR

Adds the SLF4J-JUL bridge dependency, enabling SLF4J loggers to work, and removing the logging system error on startup (http://www.slf4j.org/codes.html#StaticLoggerBinder).

### DIFF
--- a/distro/tomcat8/pom.xml
+++ b/distro/tomcat8/pom.xml
@@ -50,6 +50,12 @@
       <artifactId>apiman-distro-es</artifactId>
       <type>war</type>
     </dependency>
+
+    <!-- tomcat default logging provider for slf4j -->
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-jdk14</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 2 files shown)


outofcoffee:
@msavy Travis seems to be timing out for this build (https://travis-ci.org/apiman/apiman/builds/125297106), and one of your previous ones too: https://travis-ci.org/apiman/apiman/builds/125044833. Any ideas? :-/

msavy:
Hey Pete!

Sorry Travis is being a pain.

We've just added some logger functionality into apiman recently which allows pluggable logging impls in apiman. My plan is to do one for each platform (where appropriate). I actually completely forgot about JUL, so I'll do that as the default implementation for all platforms.

I'll wait for Eric to discuss whether we should merge this or not, as this seems related but not the same.

EricWittmann:
This seems like a reasonable change to me!  Pete should perhaps have a look at the latest logger stuff that Marc wrote and offer some feedback, and perhaps an impl for tomcat?  :)

PR #460, opened 2016-04-25
perf(gateway-engine-beans) Reimplement CaseInsensitiveStringMultiMap to substantially improve performance and reduce GC. [APIMAN-1113]

msavy:
- #toLowerCase type comparisons are now done without causing object creation.
  Caveat: map expects only ASCII for keys (as per HTTP spec), else will not
  be case insensitive.

- A fast hashing algorithm, xxHash, was selected and an implementation used
  which also causes no allocation.

- Optimised for the most common behaviour: frequent access and modifications
  with bijective mappings. Further work may be done to improve multi-value
  performance if it proves necessary.

- More efficient iterator which doesn't create anywhere near as much garbage,
  as an unmodifiable version of the real internal elements are returned.

- Iteration isn't especially efficient for very large datasets, as the entire
  underlying array is iterated over. In most cases our tables are extremely
  small, which doesn't justify the additional overhead of maintaining a
  global linked list.

- Various other work to locate and retrieve elements very quickly in the case
  of hash collisions and/or multi-values.

- Reversed the order of elements when adding multi-values. Now the last value
  in is the first value out (LIFO). This imitates the behaviour of a normal
  map more accurately. i.e. M(X, Y) = { X => [Y] }, M(X, Z) => { [Z, Y] }

- Substantially improve test coverage of CaseInsensitiveStringMultiMap given 
  custom impl.

JIRA: APIMAN-1113

### DIFF
--- a/gateway/engine/beans/pom.xml
+++ b/gateway/engine/beans/pom.xml
@@ -12,6 +12,10 @@
 
   <dependencies>
     <dependency>
+      <groupId>net.openhft</groupId>
+      <artifactId>zero-allocation-hashing</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-core</artifactId>
       <scope>provided</scope>
--- a/gateway/engine/beans/src/main/java/io/apiman/gateway/engine/beans/util/CaseInsensitiveStringMultiMap.java
+++ b/gateway/engine/beans/src/main/java/io/apiman/gateway/engine/beans/util/CaseInsensitiveStringMultiMap.java
@@ -17,43 +17,82 @@
 package io.apiman.gateway.engine.beans.util;
 
 import java.io.Serializable;
+import java.nio.ByteOrder;
 import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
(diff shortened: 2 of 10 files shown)


gateway/engine/beans/pom.xml
```diff
@@ -12,6 +12,10 @@
 
   <dependencies>
     <dependency>
+      <groupId>net.openhft</groupId>
+      <artifactId>zero-allocation-hashing</artifactId>
+    </dependency>
+    <dependency>
```
EricWittmann: I didn't realize there was a new dependency - did you check with Nick to see if this was usable?

msavy: No, I’ll check tomorrow. It’s self-contained so should be simple. Otherwise I can fall back onto something else.

msavy: Might as well ask here. @rnc mind checking whether this dep is OK and if not imported already if that's doable (there are [no transitive dependencies](https://github.com/OpenHFT/Zero-Allocation-Hashing/blob/master/pom.xml) so it should be easy peasy).

rnc: @msavy @EricWittmann  It appears to be Apache 2.0 licensed and a Maven build so its fine. Thanks for asking.

msavy: Thanks, Nick! Appreciated.

PR #461, opened 2016-04-25
Fix for API definition double save

kahboom:
Summary:
- Changed the order in which the functions are executed, since they are done so synchronously. However, for a quick second I thought I was able to replicate the issue again, but tried another 10 times to no avail. If it happens again, this could very well be a timing issue. Please verify that this fixes it on your end as well.

JIRA: https://issues.jboss.org/browse/APIMAN-1128

cc @EricWittmann

### DIFF
--- a/manager/ui/war/plugins/api-manager/css/apiman.css
+++ b/manager/ui/war/plugins/api-manager/css/apiman.css
@@ -4,6 +4,22 @@ body {
   font-size: 13px;
 }
 
+textarea#api-definition.apiman-form-data {
+  height: 100px;
+}
+
+textarea#api-definition.apiman-form-data.myFocus {
+  height: 450px;
+}
+
+textarea#api-definition.apiman-form-data.myBlur {
+  height: 100px;
+}
+
+.actions .btn:focus, .actions .btn:active:focus {
+  outline: none;
+}
+
 .ui-select-bootstrap .ui-select-choices-row>a:hover, .ui-select-bootstrap .ui-select-choices-row>a:focus {
   text-decoration: none;
   color: #262626;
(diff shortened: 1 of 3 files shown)


EricWittmann:
It looks like a function declaration/definition was moved up a bit?  How does this change the execution order of anything?

kahboom:
Summary:
- Removed reference to jQuery to control the textarea height from within the controller.
- Added ng-init (not my favorite practice, but much better than jQuery and will do for now), and ng-class to determine the textarea height.
- Adjusted the ordering of the `saveApi` method, although I think this only partially helped with the fix, as there is still a minor existing issue (see below).

@EricWittmann - This has been tested about a billion times because I couldn't get consistent results, until I finally narrowed down the issue. So, what's happening is that "where" you click on the button matters; if you click on the edge or very bottom of the button it will "focus" the button, but will not press it. I don't see this problem on other views.

I think this is a CSS-related issue, perhaps conflicting with the ngBlur and ngFocus event listeners (and respective CSS changes). I played around a bit with padding and margins to no avail without it affecting the appearance and size of the buttons. I could do that just for this page, but I didn't want to do that a) without asking, and b) because of the inconsistency. Please see the video examples below and let me know if it's worth it to continue to investigate. Otherwise, feel free to merge. With a styling and consistency re-do of the application this problem should be eliminated completely.

Replicating the issue: https://drive.google.com/open?id=0B9yOpcUD9qIadUFPLUZxUXFjZ1U

Fixed with caveat (when it works, when it doesn't): https://drive.google.com/open?id=0B9yOpcUD9qIaVUNNcm1URk1VN2M

PR #462, opened 2016-04-26
[APIMAN-1131] [APIMAN-1132] SLF4J and log4j2 logging delegates, plus refactoring.

msavy:
@EricWittmann One thing I think we should discuss is whether I should move the logging interfaces from `apiman-common-util` to `apiman-common-logging` perhaps? I'm not sure whether you can have modules and classes in the same module, though.


msavy:
@outofcoffee You should now be able to use `slf4j` or `log4j2` from the policy context.

All you need to do is ensure the appropriate classes are on the classpath and then set, for example:
 
```properties
apiman-gateway.logger-factory=io.apiman.common.logging.slf4j.Slf4jLoggerFactory
```

PR #464, opened 2016-04-27
perf(plugins) Refactoring of URL to URI to avoid URL's blocking equals and hashCode

msavy:
JIRA: APIMAN-1097


msavy:
There may be more, this was just my initial stab at the most obvious area (and the stuff connected to it).

PR #491, opened 2016-06-16
Remove unused dependencies

rnc:
@EricWittmann Can you review this please? Thanks!

### DIFF
--- a/common/logging/log4j2/pom.xml
+++ b/common/logging/log4j2/pom.xml
@@ -24,9 +24,5 @@
           <groupId>org.apache.logging.log4j</groupId>
           <artifactId>log4j-api</artifactId>
         </dependency>
-        <dependency>
-          <groupId>org.apache.logging.log4j</groupId>
-          <artifactId>log4j-core</artifactId>
-        </dependency>
     </dependencies>
 </project>
--- a/gateway/api/rest/pom.xml
+++ b/gateway/api/rest/pom.xml
@@ -33,17 +33,6 @@
       <artifactId>swagger-annotations</artifactId>
     </dependency>
     <dependency>
-      <groupId>io.swagger</groupId>
-      <artifactId>swagger-jaxrs</artifactId>
-      <exclusions>
-        <!-- avoid this old dependency which breaks CXF -->
-        <exclusion>
-          <groupId>javax.ws.rs</groupId>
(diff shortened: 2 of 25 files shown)


common/logging/log4j2/pom.xml
```diff
@@ -24,9 +24,5 @@
           <groupId>org.apache.logging.log4j</groupId>
           <artifactId>log4j-api</artifactId>
         </dependency>
-        <dependency>
-          <groupId>org.apache.logging.log4j</groupId>
-          <artifactId>log4j-core</artifactId>
```
msavy: Hmm - I think this needs to be on the classpath. Will try to confirm.

gateway/api/rest/pom.xml
```diff
@@ -33,17 +33,6 @@
       <artifactId>swagger-annotations</artifactId>
     </dependency>
     <dependency>
-      <groupId>io.swagger</groupId>
```
EricWittmann: I'm not sure about this one - I think it may be used downstream in f8.  @KurtStam - are you using this somewhere?  I *think* you added it originally, no?

KurtStam: @EricWittmann I do use the swagger editor in apiman, but I don't recollect I added it.. I can add it explicitly in ipaas if it's needed.

EricWittmann: Actually I just did a little bit more sleuthing and I think this change is OK.  The relevant commit is here:

https://github.com/apiman/apiman/commit/51fbb51c325ecfd0068853f3d516464d5cd2e8c1

You added swagger annotations to the API as well as a swagger 2.0 resource to the jax-rs layer.  The latter requires swagger-jaxrs to be added to the `gateway/platforms/war/micro/pom.xml` file, which you did.  So I think *this* reference can safely be removed, since the `gateway/api/rest/` module really only needs the swagger-annotations dependency.

common/logging/log4j2/pom.xml
EricWittmann: @msavy I think we're just waiting on confirmation from you on this one.  When you get a chance. :)

EricWittmann:
I'm going to go ahead and merge this.  If the logger dependency was removed in error, we can always add it back.

PR #493, opened 2016-06-20
Different smaller refactorings

bennetelli:
- refactored duplicated code
- extracted method to be more correct semantically


bennetelli:
you're welcome :)

PR #501, opened 2016-09-12
Use case insensitive comparison for suppressed headers.

pcornish:
# Rationale

Suppressed headers are not always removed from downstream requests to back-end APIs, nor in responses to the caller. This is because the comparison of header names is done in a _case-sensitive_ manner. According to [RFC-2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2):

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

# Background

There is a set of 'suppressed headers' in `io.apiman.gateway.platforms.servlet.connectors.HttpApiConnection`. There are two comparison loops - one for requests and one for responses. In these loops the header keys are checked against the suppressed set and conditionally included if absent.

Currently, the comparison with the suppressed set is case-sensitive, resulting in suppressed headers being erroneously included if they have the same name but different case. This is particularly problematic as it seems that at runtime, the list of header keys returned by `io.apiman.gateway.engine.beans.ApiRequest.getHeaders()` is in lowercase, meaning none of them match the suppressed set.

# Implementation

The `java.util.HashSet` used for the suppressed headers is replaced with a `java.util.TreeSet`, specifying the `String.CASE_INSENSITIVE_ORDER` comparator.

### DIFF
--- a/gateway/platforms/servlet/src/main/java/io/apiman/gateway/platforms/servlet/connectors/HttpApiConnection.java
+++ b/gateway/platforms/servlet/src/main/java/io/apiman/gateway/platforms/servlet/connectors/HttpApiConnection.java
@@ -38,11 +38,11 @@ import java.io.OutputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLEncoder;
-import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
+import java.util.List;
 import java.util.Set;
+import java.util.TreeSet;
+import java.util.Map.Entry;
 
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLSocketFactory;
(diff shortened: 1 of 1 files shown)


msavy:
We have a `io.apiman.gateway.engine.beans.util.CaseInsensitiveMultiMap` we could reuse which should be slightly more efficient. However, there are only a handful of keys  - might not be worth the effort.

msavy:
Actually, there's also already TreeSet with CaseInsensitive comparator. Hmm.

msavy:
👍

EricWittmann:
@msavy - I assume the thumbs up icon indicates your vote to accept this PR?  :)

EricWittmann:
OK, once my buddy Travis gives me the go-ahead I'll merge.  Thanks @outofcoffee !

knishantsingh:
👍

PR #509, opened 2016-11-26
Add execute blocking component

msavy:


msavy:
Will add config shortly

PR #515, opened 2016-12-13
chore(server-all) Update to WildFly 10.1.0.Final

msavy:
* Update from WildFly 10.0.0.Final to 10.1.0.Final


msavy:
(Also: Trying out the new "request reviewers" feature)

PR #517, opened 2017-01-13
Adds a Hazelcast implementation for rate limiting and shared state.

outofcoffee:
### What it is

Rate limiter and shared state implementations added for Hazelcast.

### Impl notes

* Modelled on the ISPN implementations.
* Please could you check I've included it as a dependency in the right places? I have matched where the `*-engine-ispn` dependency was used, but not sure this is right.

### DIFF
--- a/gateway/engine/hazelcast/pom.xml
+++ b/gateway/engine/hazelcast/pom.xml
@@ -1,0 +1,44 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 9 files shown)


EricWittmann:
Looks pretty reasonable on first glance.  Once travis gives the OK I'll pull it down and have a closer look.

I had a couple thoughts that may or may not be useful.  First is that you might consider creating a new one of these:

https://github.com/apiman/apiman-servers

The server configs in there are useful for testing, getting started quickly, and also for quick reference on how to configure apiman for a particular external system.

Thought number 2 was that it's also possible to create this sort of support as a plugin rather than modifying core apiman.  To be clear, modifying core apiman is *also* fine and we're happy to accept such contributions.  But if you were to prefer a plugin approach that should be possible for these components.

outofcoffee:
Thanks, @EricWittmann. Agree that things like this feel more 'plugin'-like. I must confess, I had thought https://github.com/apiman/apiman-plugins was for policy plugins only, but I see that is not the case :) Cheers again for merging this.

PR #524, opened 2017-01-29
feat(apiman-gateway) Simple HTTP backpressure first pass impl; only implemented in Vert.x platform.

msavy:


msavy:
Thinking of how to test this satisfactorily.

PR #526, opened 2017-02-01
Timeout comment in apiman.properties

ashishkp:
Timeout comment added in apiman.properties (under tomcat/wildfly distros).


EricWittmann:
Thanks. :)

PR #532, opened 2017-02-06
Add slf4j logger factory to WAR path

msavy:
@EricWittmann Is there a way to add JARs to the classpath easily in our WAR/WF distros? For instance, adding the log4j2-slf4j JAR if people wanted to use that, or logback, etc as their implementation?


ulaBLUE:
Hi, if this ok, could you please merge to branch. Need to build the deployment from master :). 
Thanks

msavy:
@ulaBLUE I *think* this will be using JBoss Logging under the covers by default :)

PR #533, opened 2017-02-06
APIMAN-1235: override maven .m2 local repository system property added

ulaBLUE:
@EricWittmann 
as discussed, I haved added below commit for APIMAN-1235.
maven .m2 local repository system property is added to the common method so it is used by both apiman-gateway and apiman-manager.


EricWittmann:
Looks good, thanks!

PR #537, opened 2017-02-08
test(registry) Build rig to allow URI pull registry to emulate push.

msavy:
After much fiddling, it does indeed work 😄!

I'll add a post later explaining how it works - as you can imagine some inventive thinking was required.


gateway/test/src/test/java/io/apiman/gateway/test/junit/GatewayRestTester.java
```diff
@@ -256,6 +256,7 @@ public void evaluate() throws Throwable {
                     }
     
                     testInfo.plan.runner.runTest(restTest, endpoint);
+                    gatewayServer.next(endpoint);
```
msavy: This additional method greatly simplified things from my initial approaches (of which I had a few); it allows you to know when the next individual test is being run, so a reload can  be issued to the registry.

gateway/test/src/test/java/io/apiman/gateway/test/junit/vertx3/ApiToFileRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+        this.eb = vertx.eventBus();
+        this.fileSystem = vertx.fileSystem();
+        linkRoot();
+        createTempFile();
+        createResetListener();
+    }
+
+    private void createResetListener() {
+        eb.consumer(ApiToFileRegistry.class.getCanonicalName(), msg -> {
+           if (msg.headers().get("action").equals("reset")) {
```
msavy: At the end of the test run, a message is sent to all `ApiToFileRegistry` instructing them to reset. It runs in a separate vert.x instance to the code under test to avoid interference.

I guess I could destroy and redeploy but this is simpler and was a result of some other attempts I had earlier (which worked all bar one test because they were hard reloading which caused metrics to be dropped).

gateway/test/src/test/java/io/apiman/gateway/test/junit/vertx3/ApiToFileRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+               throw new RuntimeException("Unknown action in " + msg.headers());
+           }
+        });
+    }
+
+    private void createTempFile()  {
+        try {
+            file = File.createTempFile("apiman", "fs");
+            System.setProperty("CONFIG_FILE_PATH", "file:///" + file.getAbsolutePath());
+            System.out.println("file path file:///" + file.getAbsolutePath());
```
msavy: Set the path in System properties so when the registry under test boots up it will know where the file we've written is (remember, it's in a different Vert.x instance - but in the same JVM).

gateway/test/src/test/java/io/apiman/gateway/test/junit/vertx3/ApiToFileRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+            file.deleteOnExit();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public void reset() {
+        fileSystem.deleteBlocking(file.getAbsolutePath());
```
msavy: While we delete the file on reset, we must keep the original file reference because if we do a reload of the polling registry (rather than a restart) it will still be using the existing URI - and hence if we change to a new file here, the polling registry would just keep loading the old version.

gateway/test/src/test/java/io/apiman/gateway/test/junit/vertx3/ApiToFileRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+        linkRoot();
+    }
+
+    private void rewrite() {
+        fileSystem.writeFileBlocking(file.getAbsolutePath(), Buffer.buffer(root.encodePrettily()));
+        restartReader();
+    }
+
+    private void restartReader() {
+        eb.publish("restartReader", "reset!");
```
msavy: Can remove this now, I think.

gateway/test/src/test/java/io/apiman/gateway/test/junit/vertx3/ApiToFileRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private void publish(Api api) {
+        System.out.println("WANT TO PUBLISH " + api);
```
msavy: Remove

gateway/test/src/test/java/io/apiman/gateway/test/junit/vertx3/ApiToFileRegistry.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private void retire(Api api) {
+        JsonObject removeJson = apiMap.remove(api);
+        apis.remove(removeJson);
+        rewrite();
+    }
+
+    private void register(Client client) {
+        JsonObject clientJson = new JsonObject(Json.encode(client));
```
msavy: These methods use the InMemoryRegistry to do validation, and then rewrite the JSON file when necessary.

gateway/test/src/test/java/io/apiman/gateway/test/junit/vertx3/Vertx3GatewayFileRegistryServer.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * A Vert.x 3 version of the gateway test server
+ *
+ * @author Marc Savy {@literal <msavy@redhat.com>}
+ */
+@SuppressWarnings("nls")
+public class Vertx3GatewayFileRegistryServer implements IGatewayTestServer {
+
+    protected static final int API_PORT = 9009;
```
msavy: Note the trick here, the port reported to the test framework matches that defined for the emulating registry (`ApiToFileRegistry.java`), however the port given to the registry under test is different - so its methods are never invoked (which makes sense as it's a polling/pull registry).

Somewhat sleight of hand, but works very nicely.

gateway/test/src/test/java/io/apiman/gateway/test/junit/vertx3/Vertx3GatewayFileRegistryServer.java
```diff
@@ -0,0 +1,162 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public void next(String endpoint) {
+        try {
+            if (URI.create(endpoint).getPort() == API_PORT) {
+                CountDownLatch cdl = new CountDownLatch(1);
+                URILoadingRegistry.reloadData(done -> {
```
msavy: Had to add some extra functionality to the registry to force to reload. This is likely to be useful anyway as a true polling/pull registry rather than a one-shot immutable pull registry.

gateway/test/src/test/resources/vertx3/conf-api-to-file.json
```diff
@@ -0,0 +1,139 @@
...
+    "verticles": {
+        "http": {
+            "port": 8999,
+            "count": 0
+        },
+        "https": {
+            "port": 8998,
+            "count": 0
+        },
+        "api": { "port": 9009, "count": 1 }
```
msavy: Important part.

gateway/test/src/test/resources/vertx3/conf-file.json
```diff
@@ -0,0 +1,135 @@
...
+    "http": {
+      "port": 8082,
+      "count": 1
+    },
+    "https": {
+      "port": 8443,
+      "count": 0
+    },
+    "api": {
+      "port": 8081,
```
msavy: Nothing will invoke this during the tests.

gateway/test/src/test/java/io/apiman/gateway/test/junit/vertx3/Vertx3GatewayFileRegistryServer.java
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright 2015 JBoss Inc
```
msavy: Wrong date

PR #545, opened 2017-02-14
Change ESMetrics to use batching

EricWittmann:


gateway/engine/es/src/main/java/io/apiman/gateway/engine/es/ESMetrics.java
```diff
@@ -96,13 +110,22 @@ public void run() {
...
-            Index index = new Index.Builder(metric).refresh(false)
-                    .index(getIndexName())
-                    .type("request").build(); //$NON-NLS-1$
-            DocumentResult result = getClient().execute(index);
+            Collection<RequestMetric> batch = new ArrayList<>(this.batchSize);
+            RequestMetric rm = queue.take();
+            batch.add(rm);
+			queue.drainTo(batch, this.batchSize - 1);
+			
+			Builder builder = new Bulk.Builder();
```
msavy: What happens if you're below the batch threshold  but no new stuff comes? Is there a timer to also flush it out? e.g. after 5 minutes. I made a timer component you could use if that seems useful.

msavy: Not necessarily required, just a thought.

msavy: https://github.com/apiman/apiman/blob/master/gateway/platforms/servlet/src/main/java/io/apiman/gateway/platforms/servlet/components/PeriodicComponentImpl.java

EricWittmann: The call to `take()` blocks until something is ready.  Then the call to `drainTo()` returns AT MOST `this.batchSize - 1` items.  But it doesn't block.  So if nothing is in the queue, then it returns immediately.  So the thread is NEVER blocked waiting for the queue to fill up to a minimum size.

msavy: Gotcha, I see. It always runs as quickly as it can, and sends a batch of up-to a given size (rather than waiting for the size to accumulate to a specific target and/or time).

Sounds good, thanks!

EricWittmann: Just so! :)

msavy:
Thanks!

PR #550, opened 2017-02-18
build(manager-ui) Use Yarn instead of NPM.

msavy:
Faster builds, more deterministic dependency resolution, and better caching.


msavy:
@EricWittmann any objections? Cuts a nice slice off CI time.

msavy:
On second thoughts, will merge and happy to revert/unpick if you have any issues with it.

msavy:
BTW apiman/apiman-studio#18 might be resolved using this same approach. Adding this comment to link it in.

EricWittmann:
Looks good to me!

PR #551, opened 2017-02-18
build(platform-vertx) Fix shade plugin build warnings. Upgrade to 2.4.3.

msavy:
Sorry, another small warning-eliminating build PR :-).


msavy:
Warnings were

```
[WARNING] Map in class org.apache.maven.plugins.shade.resource.ManifestResourceTransformer declares value type as: class java.util.jar.Attributes but saw: class java.lang.String at runtime
[WARNING] Map in class org.apache.maven.plugins.shade.resource.ManifestResourceTransformer declares value type as: class java.util.jar.Attributes but saw: class java.lang.String at runtime
```

PR #555, opened 2017-03-07
3scale pull registry (and other associated enhancements)

msavy:
It's a big blob of _stuff_ at the moment as it's the result of extended work between this, the 3scale plugin, enhancements to facilitate both (and a bunch of other stuff)  - I'll break it into more logical commits later. 

- [x] Add Javadoc where necessary.
- [ ] Update Vert.x README again (although need to consider just making proper web content for it, as it's too much for a README, likely).
- [x] Ensure copyright and class headers added.
- [x] That thing you forgot.
- [x] Check for any dev logging that should be removed (or added!).
- [x] Consider whether some of the functionality I ended up putting in the bean(s) should be elsewhere? Regex route-matching for instance. 
- [ ] Address TODO/FIXME points.


gateway/engine/beans/src/main/java/io/apiman/gateway/engine/beans/Api.java
```diff
@@ -46,6 +46,8 @@
     @JsonIgnore
     private transient long maxPayloadBufferSize;
     private List<Policy> apiPolicies = new ArrayList<>();
+    @JsonIgnore
+    private transient Map<String, Object> cache = new HashMap<>();
```
msavy: This is from an older revision, don't need it anymore. Kill.

msavy:
Going to commit this now, I'll do the refactoring for the servlet impl in a separate PR. Don't want to leave this hanging off for ages.

PR #593, opened 2017-04-27
fix(gateway-vertx) Ensure #resume called in error cases before #end to avoid potential TCP hang.

msavy:
Thanks to @EricWittmann for helping me discover this one. I was aided in fixing the issue by @vietj's comment @ https://github.com/eclipse/vert.x/issues/1244#issuecomment-165532673.


EricWittmann:
Perhaps you need to rename #resume to #cv ?

PR #607, opened 2017-06-02
build(manager-api-micro): Missing ICU dep for micro

msavy:
@EricWittmann I think this issue was caused when you added something around i18n to WAR but it wasn't also added to Micro. Not sure if I need to add it anywhere else? 

It seems to cause some failures in certain parts of the UI when using any of the micro distros, anyway.


msavy:
This is with apiman-servers though, so not sure whether that is different.

msavy:
Received no objections, so going to merge.

PR #626, opened 2017-08-17
Update SSLSessionStrategy for devMode to allow all TLS version

alesuffleur:
The previous code allow only TLS v1.0 when devMode is true.

With this modification, all versions of TLS will be accepted.


msavy:
Thanks for your contribution @alesuffleur

PR #630, opened 2017-10-09
build(distro-wildfly10) Ensure archiveBaseDirectory is set.

msavy:


msavy:
I think Java 1.8.0_144 might have inflated memory usage

PR #634, opened 2017-11-11
feat(gateway-api) List entities on gateway API (et al)

msavy:
* feat(gateway): When copying buffer, set size hint for new Vert.x/Netty buffer.

* feat(gateway-api) Indicate that Gateway API filters are compatible with async.

* feat(platform-vertx): Rewrite Gateway API to use Resteasy-Vertx for JAX-RS.

* fix(engine-es): Better escaping when entity is not arriving via JSON.

* build(pom): Update RESTEasy JAX-RS to 3.1.4.Final


msavy:
This has been in the works for a while. As it has been rebased and squashed multiple times, the dates on the PR/commit are misleading.

msavy:
Will add some explicit tests for new features shortly.

PR #639, opened 2017-11-27
Fixes broken ES cache population.

outofcoffee:
### Background

Cache population for ES doesn't work for the simple/non-binary implementation.

### This PR

Ensures the JSON object is set in the `head`, instead of the cache entry itself(!) in `ESCacheStoreComponent.put()`, from where it is expected to be read in `ESCacheStoreComponent.get()`.


msavy:
Oh goodness, that's a bit of a clanger. This looks like a good fix to me!

PR #641, opened 2017-11-28
Allows Prometheus metrics implementation to be used with no client ID

outofcoffee:
### Background

The Prometheus metrics implementation throws an exception when `null` is passed as the client ID. This is because client ID is a mandatory label, and labels can't be `null`.

### This PR

* Allows Prometheus metrics implementation to be used with no client ID.
* Reenables the Prometheus metrics tests (which start their own server).


outofcoffee:
Cheers, @msavy. Style comment noted :-)

PR #647, opened 2017-12-01
build(pom) Update to Netty 4.1.17.Final

msavy:


msavy:
Going to leave this until we update Vert.x itself as it seems to have some weird side-effects.

PR #649, opened 2017-12-06
Prevents the root cause for storage exceptions being swallowed

outofcoffee:
## Overview

Prevents the root cause for storage exceptions being swallowed by a failure to roll back the transaction.

## Issue

When certain storage exceptions occur, such as a failure to log in to the database, apiman attempts to roll back the transaction. If the transaction rollback fails (or the transaction is not active), then another exception is thrown. Unfortunately, this second exception masks the first - so the original cause is lost, making debugging very difficult.

## This PR

Ensures the original storage exception is thrown, optionally adding any rollback exceptions to the 'suppressed' list.


msavy:
@outofcoffee I had to revert this temporarily because it was causing issues (which were not covered by tests in apiman/apiman).

For example:

`curl -kv 'http://admin:admin123!@localhost:8080/apiman/gateways/some-gw-that-does-not-exist'` would return a 500 error instead of 404.

PR #650, opened 2018-02-21
Fix example configuration with comma separated list

alesuffleur:
A comma  separated list work only if the comma is escaped. Because the communs-configuration library (version 1.10) is called with getString(), then that return only the first word.
With the value allowedProtocols=TLSv1.2,TLSv1.1, apiman load only "TLSv1.2".

To fix issues we can:
- Escaped comma,
- Move library  communs-configuration to 2.2 ( more complicated)
- Use getArrayString(), many refactoring because getString is use in generic code for String and List


msavy:
Thanks for this information. Will try to investigate and fix this. It is a bit strange, you're right.

I tried to move to commons config 2.2 in the past but it was actually a lot more work than it appeared.

PR #651, opened 2018-02-27
Fix cache with different query params

alesuffleur:
The buildCacheID method ignored the query params.
I added a md5 of query params.

Maybe it's better to make a hash of rawrequest to cache POST/ PUT/PATCH verb.


alesuffleur:
I will work and add some unit test.

PR #652, opened 2018-02-28
Fix cache with different query params

alesuffleur:
The buildCacheID method ignored the query params.
I added a md5 of query params.

Maybe it's better to make a hash of rawrequest to cache POST/ PUT/PATCH verb.

### DIFF
--- a/gateway/engine/policies/src/main/java/io/apiman/gateway/engine/policies/CachingPolicy.java
+++ b/gateway/engine/policies/src/main/java/io/apiman/gateway/engine/policies/CachingPolicy.java
@@ -171,7 +171,6 @@ public class CachingPolicy extends AbstractMappedDataPolicy<CachingConfig> imple
         }
     }
 
-<<<<<<< Upstream, based on masterApiman
 	/**
 	 * Builds a cached request id composed by the API key followed by the HTTP verb
 	 * and the destination. In the case where there's no API key the ID will contain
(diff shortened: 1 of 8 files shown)


bekihm:
What is the status about this Pull-Request? Otherwise we cannot use the Caching-Policy.

msavy:
I haven't had time to work with it yet (probably needs an additional option in the UI also). 

Unfortunately, I'm not able to spend a lot of time on this at the moment, but I'll try to get around to it shortly.

What's your use-case? Do you just cache a response if all query parameters are the same, or do you need to be able to pay attention to a subset?

alesuffleur:
> "What's your use-case? Do you just cache a response if all query parameters are the same, or do you need to be able to pay attention to a subset?"


For example, when you have a web service with pagination:
- http://mygateway/webservice/documentation?page=1
- http://mygateway/webservice/documentation?startItem=50

msavy:
Sure, but I just mean in the case when you haves several query parameters, e.g.: ?a=123&unimportant=foo

Would one expect this to be configurable so that (e.g.) `a=123` is part of the cache key, but `unimportant=foo` is not?

msavy:
Thanks, I understand. 

I’ll try to work with it shortly to integrate the feature.

Apologies for not being as active any more — just an issue of my employer wanting me to work primarily on another area at the moment.

bekihm:
We can also offer to test this Pull-Request and extend it for our purpose described above.

bekihm:
I think this PR is now closed by [PR 712](https://github.com/apiman/apiman/pull/712).

PR #656, opened 2018-04-11
Fixes plugin logging in Tomcat distributions.

pcornish:
Following on from https://github.com/apiman/apiman/pull/654, this PR adds the SLF4J API dependency to Tomcat distributions to ensure logs from plugins are also routed correctly.


pcornish:
Thanks, @msavy.

PR #657, opened 2018-04-12
feat(gateway-servlet): Allow HTTP/1.0 when connecting to backend.

msavy:
APIMAN-1295


msavy:
Argh, it looks like OkHttp doesn't support HTTP 1.0 properly, sorry .

I'll look into whether it'll be possible by changing versions, but doesn't look hopeful

PR #660, opened 2018-05-17
apiman-1318: Export/Import of a plugin with policies does not work because of its ID

volkflo:
See also: https://issues.jboss.org/projects/APIMAN/issues/APIMAN-1318

This may be not the ultimate solution but it is working for us at the moment.

Please let me know what you think.


msavy:
Eric and I discussed this and we think it looks good. 

Have you done any testing on this? Have you noticed any issues?

volkflo:
Thanks for your answer Marc.

Yes I tested this with some configurations via Export/Import and also as basic configuration at the first startup of apiman in tomcat. This is working for me.
Unfortunately, I could only test with elasticsearch and not with any database.

The tests are all good on windows and linux, but one of your travis-tests failed. I don't checked why.

msavy:
Thanks for your patience. I've finally slayed all of the bugs (hopefully) from the ES 5.x upgrade. 

I'll asses, merge, and test your work after that's on master.

msavy:
Thank you for your contribution, it's much appreciated 🎉

PR #668, opened 2018-06-12
Improve wildcard ("*") search in UI

volkflo:
Apiman allows to find all entries such as all APIs, Orgs, User per wildcard.
But the results are wrong because there is a paging. So the return is limeted to 20 or 50 entries. 
So I make a suggestion to change this because otherwise this is a bit confusing.

I created two screenshots that you can see what I mean:

Before:
before

After:
after

Please let me know what you think :)

### DIFF
--- a/manager/ui/war/plugins/api-manager/ts/consumer/consumer-apis.ts
+++ b/manager/ui/war/plugins/api-manager/ts/consumer/consumer-apis.ts
@@ -19,6 +19,8 @@ module Apiman {
                         var body:any = {};
                         body.filters = [];
                         body.filters.push( {"name": "name", "value": "*" + params.q + "*", "operator": "like"});
+                        body.page = 1;
+                        body.pageSize = 0x7fffffff; // Java Interger.MAX_VALUE
                         var searchStr = angular.toJson(body);
                         
                         ApimanSvcs.save({ entityType: 'search', secondaryType: 'apis' }, searchStr, function(reply) {
--- a/manager/ui/war/plugins/api-manager/ts/consumer/consumer-apis.ts
+++ b/manager/ui/war/plugins/api-manager/ts/consumer/consumer-apis.ts
@@ -20,7 +20,7 @@ module Apiman {
                         body.filters = [];
                         body.filters.push( {"name": "name", "value": "*" + params.q + "*", "operator": "like"});
                         body.page = 1;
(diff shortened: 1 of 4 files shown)


volkflo:
This will not work with Elasticsearch 5.X, <ou will get an exception in UI due to this ES-Error:

`{"error":{"root_cause":[{"type":"query_phase_execution_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [2147483647]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"apiman_manager","node":"rRcmQ9fYQAeUfk4urdhzGQ","reason":{"type":"query_phase_execution_exception","reason":"Result window is too large, from + size must be less than or equal to: [10000] but was [2147483647]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level setting."}}]},"status":500}`

I created a fixup-commit for you and changed the pageSize to 10000.
I think 10000 Orgs/APIs/Users will also fit for the most usecases.

msavy:
Could you please respond to my comment on your code? Are you changing this to try to return everything at once in the query instead of paginating?

Not sure that's a great approach.

Is there nothing in the API at the moment that indicates the record count accurately? In that case, I should add something either to the returned data or a new API method.

volkflo:
Sorry for not responding, I've been on vacation the last days. But where can I find your comments? I did not get any notification about this :/

Nevertheless you got my idea (or better my workaround): My problem is that I dont know all names of the APIs or Orgs so I use in the search the wildchard (*) to get all of the entries at once in the UI (on Find APIs, Find Orgs, Contracts and Catalog Importer (Search for API) and to find members).

On this pages in UI the page size is limited to 20 or 50 and there is no pagination implemented. This is only implemented on the change log page ('Show next 20 items')

So I use the the maximum value for the moment as workaround to get all entries at once.

For me a better approach for this pages would be the same as for the catalog, we show always everything and then put a filter on this. The way that you can only find something if you know the name is quite hard if you have a lot of entries.

tevosouza:
I think that is resolved on pr #810 and #815 maybe this pr needs to be closed.

bekihm:
@tevosouza You are right we waited until all your PRs were done :-)

PR #670, opened 2018-06-15
refactor(engine-es): Update index initialisation error suppression

msavy:


msavy:
Please do feel free to review after the fact and I'll add any corrections.

PR #686, opened 2018-07-04
Allow policies to process failures

msavy:
Prototype for people on the ML to see. Merged version will have tests and fuller documentation


msavy:
Test and docs added.

PR #706, opened 2018-08-29
APIMAN-1339: Support self signed certificates from keycloak in vertx

volkflo:
This is a fix for https://issues.jboss.org/browse/APIMAN-1339

Without this fix it is not possible to connect vertx with a standalone keycloak installation that is secured by a self signed certificate.

Please let me know what you think :)

### DIFF
--- a/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/api/auth/KeycloakOAuthFactory.java
+++ b/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/api/auth/KeycloakOAuthFactory.java
@@ -20,9 +20,11 @@ import io.apiman.common.util.Basic;
 import io.apiman.gateway.platforms.vertx3.common.config.VertxEngineConfig;
 import io.apiman.gateway.platforms.vertx3.verticles.ApiVerticle;
 import io.vertx.core.Vertx;
+import io.vertx.core.http.HttpClientOptions;
 import io.vertx.core.json.JsonObject;
 import io.vertx.core.logging.Logger;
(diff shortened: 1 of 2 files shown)


gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/api/auth/KeycloakOAuthFactory.java
```diff
@@ -143,8 +148,42 @@ public AuthHandler addAuthorities(Set<String> authorities) {
...
+        if (config.containsKey("allow-any-hostname")) {
+            Object value = config.getValue("allow-any-hostname");
+            if (value.equals(true) || value.equals("true")) {
+                options.setVerifyHost(false);
+            }
+        }
+
+        if (config.containsKey("disable-trust-manager")) {
+            Object value = config.getValue("disable-trust-manager");
+            if (value.equals(true) || value.equals("true")) {
```
volkflo: With `if (value.equals(true) || value.equals("true")) ` you are able to use System Properties in the conf-X.json like so:

```
"disable-trust-manager": "${devMode}",
"allow-any-hostname" : "${allowAnyHost}"
```

EricWittmann:
@msavy any problems with this?

volkflo:
@EricWittmann What is the status of this? Would be very nice if we could include this :)

msavy:
I'm just slightly concerned this might break things for something other than Keycloak, but otherwise I'm happy for this to be merged

EricWittmann:
Rolling the dice!  :)

volkflo:
Thanks for merging, if something goes wrong with providers other than keycloak, you can blame me for this :)

PR #708, opened 2018-08-29
fix-ui: Show selected tags in API Catalog

volkflo:
This is a little UI fix for the API Catalog. 
If you use the filter for tags it gets annoying if the selected tags are not shown.

## Before
before

## After
after


EricWittmann:
This looks like a nice change, any reason not to merge this, @msavy ?

msavy:
Lgtm

volkflo:
@EricWittmann and @msavy Thanks again for merging this :)

PR #709, opened 2018-09-13
[FIX #705]Add PATCH verb

### LINKED ISSUE #705: APIMan does not handle PATCH verb on Authorization Policy
While trying to setup some policies on our webservices through apiman, we found out that it does not seem to handle PATCH method (ie link).
I tried to dig in documentation but there does not seem to be anything about that. 
Is there a workaround for this, like using POST mapping or else or am i missing something ?
apiman

alesuffleur:
Add PATCH http verb to requestMethod using in authorization policy and Ignored Resources Policy


msavy:
Does this work without needing to modify any code in the policies you mentioned?

alesuffleur:
Yes, it work without needing to modify any code in the policies.
 It only need declaration via GUI, the policy already have compatible code

PR #712, opened 2018-12-04
Make Caching policy more RFC7234 compliant w.r.t. query string

outofcoffee:
## What is this?
* An alternative to #652, with a configuration option to maintain backwards compatibility.
* This enables the CachingPolicy to optionally include query string in its cache key, per RFC7234.

## Why?

According to RFC7234 (https://tools.ietf.org/html/rfc7234#section-2),
> 'The primary cache key consists of the request method and target URI.'

For historical reasons, this is not the behaviour of this policy, which instead uses only the path part of the URI, not the query string. The behaviour mentioned in the RFC can be enabled using a configuration option, `includeQueryInKey`, whose default is `false`.

Appropriate test coverage is added to verify the behaviour with the configuration [en|dis]abled. Also reenables ignored unit tests for QueryMap and adds another test case for an empty map.

## Implementation considerations

We should consider whether to default the behaviour to `true` as it currently violates the principle of least surprise (see #652).

### DIFF
--- a/distro/data/src/main/resources/data/all-policyDefs.json
+++ b/distro/data/src/main/resources/data/all-policyDefs.json
@@ -99,7 +99,7 @@
       "templates" : [
         {
           "language" : null,
-          "template" : "API responses will be cached for @{ttl} seconds."
+          "template" : "API responses will be cached for @{ttl} seconds. @if{includeQueryInKey == true}The query string will be included for the cache key.@end{}@if{includeQueryInKey == false}The query string will be ignored for the cache key.@end{}"
         }
       ]
     },
--- a/gateway/engine/beans/src/test/java/io/apiman/gateway/engine/beans/util/QueryMapTest.java
+++ b/gateway/engine/beans/src/test/java/io/apiman/gateway/engine/beans/util/QueryMapTest.java
@@ -16,33 +16,35 @@
 
 package io.apiman.gateway.engine.beans.util;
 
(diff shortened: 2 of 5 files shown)


EricWittmann:
Is there any discussion needed on this before merging?  I see that it's pretty darn old but without any objections or comments.  @msavy ?

bekihm:
The code looks good so far. Is there also the possibilty to take care about no-cache header? And if the request is for example a POST request with a body is there a way to cache POST requests which only differ on different bodies (for example hashing the body and add it to the caching key)?

EricWittmann:
@bekihm I think those sound like good enhancements.  Any chance you're interested in submitting a PR for those?  :)

bekihm:
@EricWittmann We from @e2ebridge will try to do our best to make some good PR in the future.

bekihm:
@EricWittmann Just a question about the policy configuration. How can I enable the new 'includeQueryInKey' configuration in production? In the test case it is done via method annotation but for production it must be another way.

EricWittmann:
That test annotation is just a way to get standard policy configuration into the test framework without having to externalize it into a separate file.  In a normal running system, the `includeQueryInKey` property is simply part of the JSON configuration of the policy, like all other policy config properties.  I hope that makes sense!  

In other words, that property must be enabled on the configuration of the policy itself, and cannot be set globally.

PR #714, opened 2018-12-12
Small fix for redirect in Tomcat (with Keycloak)

volkflo:
If the session was not valid there was no redirect to the login page of Tomcat 8.
With this small fix we are able to do a redirect and have a whitelisting of the login CSS files.

This works also if you use keycloak as login provider instead of tomcat. Otherwise also the keycloak login page was not loaded correctly.

Thanks @bekihm for finding this out and help me with the debugging of this :)

@msavy and @EricWittmann Please let us know what you think of this.

### DIFF
--- a/manager/ui/war/tomcat8/src/main/webapp/login.html
+++ b/manager/ui/war/tomcat8/src/main/webapp/login.html
@@ -5,6 +5,7 @@
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
             <meta name="viewport" content="width=device-width,initial-scale=1"/>
     <title>Log in to apiman</title>
+    <base href="/apimanui/"/>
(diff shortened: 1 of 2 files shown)


manager/ui/war/tomcat8/src/main/webapp/login.html
```diff
@@ -5,6 +5,7 @@
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
             <meta name="viewport" content="width=device-width,initial-scale=1"/>
     <title>Log in to apiman</title>
+    <base href="/apimanui/"/>
```
volkflo: Needed for loading CSS if you are not on the login.html page

volkflo:
Thx for the fast merge :)

PR #718, opened 2019-01-22
Fix how vertx public endpoint is assembled in UI

volkflo:
Hi togehter,

we have encountered a small issue with the vertx json config:
The problem is how the endpoint was builded if you set a special endpoint and use port 443 or 80, but not specify this port in the endpoint sring:

**Example**:
```json
// You can force a particular endpoint to be reported here (e.g.
  // if you have some clustered setup with exotic DNS setup)
  "endpoint": "https://gateway.e2e.ch",

  // Verticle configuration
  // Port - The port a given verticle listens on (where relevant)
  // Count - Number of given verticle type launched
  "verticles": {
    "http": {
      "port": 8082,
      "count": 1
    },
    "https": {
      "port": 443,
      "count": 0
    },
    "api": {
      "port": 8081,
      "count": 1
    }
  },
```

**Result:**
This ended up in the UI with a missing slash after the host under "Endpoint"
* <https://gateway.e2e.chOrg/API/Version>
Instead of:
* <https://gateway.e2e.ch/Org/API/Version>

We refactored the assembly of the endpoint a little bit to make it more robust and also added some simple test cases.

Please let us know what you think :)

Regards
Florian

P.S. Thanks again to @bekihm for reviewing and testing this with me.

### DIFF
--- a/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/api/ApiResourceImpl.java
+++ b/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/api/ApiResourceImpl.java
@@ -16,7 +16,6 @@
 
 package io.apiman.gateway.platforms.vertx3.api;
 
-import io.apiman.common.util.SimpleStringUtils;
 import io.apiman.gateway.api.rest.contract.IApiResource;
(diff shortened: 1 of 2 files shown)


EricWittmann:
Nice changes, thanks.  In particular I'm always happy to see additional unit tests!  :)

PR #723, opened 2019-03-27
Fix InMemoryRateLimiter

volkflo:
We have detected a bug in the InMemoryRateLimter while using it with vertx.

Because vertx creates multiple threads we have in each thread a different buckets Map containing different entries for the same bucketId.

You can see this here, that the same request is handled by two different threads and this results in a wrong remaining limit:
```
2019-03-27T13:33:13.631314600Z [DEBUG] 2019-03-27 13:33:13.630 [vert.x-eventloop-thread-2] HttpConnector - Connecting to petstore.swagger.io | ssl?: true port: 443 verb: GET path: /v2
...
2019-03-27T13:34:04.557650800Z [DEBUG] 2019-03-27 13:34:04.557 [vert.x-eventloop-thread-3] HttpConnector - Connecting to petstore.swagger.io | ssl?: true port: 443 verb: POST path: /v2
```

If we make the the Hashmap static we can avoid this problem.
@EricWittmann Do you have an idea to test this. We are not sure if we can easyily test this with the existing Mockito tests.

And of course thanks @bekihm for helping me to resolve this bug :)

### DIFF
--- a/gateway/engine/core/src/main/java/io/apiman/gateway/engine/impl/InMemoryRateLimiterComponent.java
+++ b/gateway/engine/core/src/main/java/io/apiman/gateway/engine/impl/InMemoryRateLimiterComponent.java
@@ -33,7 +33,7 @@ import java.util.Map;
  */
 public class InMemoryRateLimiterComponent implements IRateLimiterComponent {
 
-    private Map<String, RateLimiterBucket> buckets = new HashMap<>();
+    private static Map<String, RateLimiterBucket> buckets = new HashMap<>();
 
     /**
      * Constructor.


EricWittmann:
I'm guessing there may be a more vert.x-y way to solve this... @msavy - as you are master of vert.x do you want to chime in on this?

As for testing, I'm not sure we can design a mockito test for this, but perhaps an integration test makes sense.  Let me give it some thought.  I'm out at a conference this week but will try to circle back around on this when I get back.

bekihm:
@msavy Instead of using the event bus of vertx. I thought about using vertx shared data api:
https://vertx.io/docs/vertx-core/java/#_using_the_shareddata_api

Because we are encounter now the problem that the InMemoryCacheStoreComponent class is not shared between different vertx verticle threads.

PR #734, opened 2019-05-28
Make TLS protocols configurable for vert.x (Client <-> Vert.x Server)

volkflo:
We had the requirement that the the vert.x gateway (as server) defines the TLS protocol version for the connection between Client <-> Gateway.

This was not possible with the existing configuration.

I had to upgrade vert.x to get some methods to achieve this functionality and add a new flag to the configuration file. 
The other setting in `connector-factory` section is only for the connection Gateway <-> Unmanaged API. 

So this was not suitable for our case and for vert.x it is only a dummy setting. 
Vertx will only log: "Can't set allowed protocols on Vert.x gateway"

So I put the new setting in the `ssl` section of the config file.

I think we have to change the documentation also a bit for this, or the other conf.json files. I will do this if this gets merged.

And again I have to thank @bekihm for testing this.

### DIFF
--- a/distro/vertx/src/main/resources/overlay/configs/conf-es.json
+++ b/distro/vertx/src/main/resources/overlay/configs/conf-es.json
@@ -244,17 +244,20 @@
     }
   },
 
-  // SSL configuration to the gateway's *front end* (i.e. client <-> gateway).
-  //  "ssl": {
-  //    "keystore": {
-  //      "path": "/the/keystore/path/here.jks",
-  //      "password": "password-here"
-  //    },
-  //    "truststore": {
-  //      "path": "/the/truststore/path/here.jks",
-  //      "password": "password-here"
-  //    }
-  //  },
+//   SSL configuration to the gateway's *front end* (i.e. client <-> gateway).
+//    "ssl": {
+//      "keystore": {
+//        "path": "/the/keystore/path/here.jks",
+//        "password": "password-here"
+//      },
+//      "truststore": {
+//        "path": "/the/truststore/path/here.jks",
+//        "password": "password-here"
+//      }
+//    },
+//    Allowed TLS/SSL protocols for Client <-> Gateway (Server)
+//    "allowedProtocols": "TLSv1.1, TLSv1.2"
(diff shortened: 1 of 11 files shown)


msavy:
This seems fine, as long as you update the docs and examples as mentioned.

Also, updating the version of Vert.x might break the OAuth auth functionality on the Gateway API, so double check that if possible.

volkflo:
I am not totally sure if I exactly know what you mean with OAuth functionality of the gateway.
Do you mean the username and password login from the UI in "Manage Gateways"?

msavy:
For some people who were using Apiman without a UI, we added the ability to
authenticate with the Gateway API using OAuth (tested with Keycloak).

Not sure how many people are still using it!

https://github.com/apiman/apiman/blob/master/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/api/auth/KeycloakOAuthFactory.java

https://github.com/apiman/apiman/blob/master/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/api/auth/AuthFactory.java#L54

Meant that you could use OAuth to speak to the Gateway API instead of
BASIC, for example.

I'll have a look to see if there are some docs. I seem to remember writing
some a while ago...

volkflo:
Ah, I see. 
I have tested this with the following auth configuration.
```json
// Gateway API Authentication. See documentation for further possibilities..
  "auth": {
    "type": "keycloak",
    "config": {
      "flowType": "PASSWORD",
      "requiredRole": "realm:apipublisher",
```
This calls `doBasic2Oauth`. But I also need a username and password for this. 
2019-05-28 15_05_19-Postman

If I change the flowType to "AUTH_CODE" then I can call the API directly with the OAuth Token from Keycloak.
2019-05-28 15_31_12-Postman

Is this what you mean?

volkflo:
I rebased this to resolve the conflicts and opened a PR for the installation guide (https://github.com/apiman/apiman-docs-installation-guide/pull/3)

PR #742, opened 2019-06-24
Fixes #741 - Broken gateway error messages

### LINKED ISSUE #741: Broken gateway error messages
We receiving the the following error message from the gateway
> Can't find bundle for base name io.apiman.gateway.engine.es.i18n.messages, locale en_US

instead of:

> No client found for API Key 7000cd57-fdef-4dfc-9971-6823504ef86e

This is causes through the refactoring we made in PR https://github.com/apiman/apiman/pull/730.

We moved `messages.properties` also with the other stuff in the`common` folder, but we need the `messages.properties` at the old directory location.

I will provide a little PR in the next days for this :)

volkflo:
I just moved the missing file back to its old location.


volkflo:
Thanks for fast merging 🎉

Issue #737, opened 2019-06-04
Improve Caching Policy including HTTP method and path pattern


bekihm:
Hello @EricWittmann ,
We have a first in progress prototype for a caching policy which meets our requirements.  You can restrict the caching with this policy to some path resources, http methods and status codes. Unlike the old caching policy, which caches all replies from the time the policy was created, no caching is performed by default.
The old policy cached all responses by default without care about http methods and path resources.  We can use wildcards for http methods and path resources to support the old behavior.
![2019-06-04 10_33_15-API Management - Edit Policy](https://user-images.githubusercontent.com/33115480/58869192-08bca600-86be-11e9-9062-613904843844.png)

We are looking forward to your opinion.
Kind regards
Benjamin Kihm

EricWittmann:
These look like nice additions to caching - did you create your own separate custom policy, or did you modify the existing caching policy?

bekihm:
In fact, I have adapted the existing one.
That means we have modified the CachingConfig class which is serialized. So there is the need to migrate already existing data of the old policy.

EricWittmann:
In that case it would perhaps be easier to duplicate the caching policy instead?

bekihm:
We thought about it. The migration of the old caching policy to our new one would be ugly to implement. The new policy works also very different to the old caching. So we will deliver it as a new policy named "Caching Resources" and may mark the old caching policy as deprecated.

bekihm:
@EricWittmann Today I will prepare a PR to cache POST requests concerning their payload (https://github.com/apiman/apiman/pull/746#issuecomment-513747780). I think then we can close this issue.

EricWittmann:
Now that your PR was merged, should this be closed?

PR #746, opened 2019-07-19
Caching resources policy

bekihm:
We present now the PR for the Caching Resources Policy like already announced in this [issue](https://github.com/apiman/apiman/issues/737 ).
You can cache requests based on their URL path, http method and specific status code.
2019-07-18 15_37_35-API Management - Add Policy

We allow also wildcards to support the cache-all functionality of the old Caching Policy. So we marked the Caching Policy as deprecated:
2019-07-17 11_02_45-API Management - Add Policy

There are also new test cases for the policy to ensure that caching works in all cases. The documentation will follow in case you accept the new policy.

Special thanks to @volkflo for reviewing and testing this new policy multiple times.

We hope you will enjoy this new policy.


EricWittmann:
I'll merge this on 2019-07-24 if there are no objections before then.

bekihm:
@EricWittmann I will add another feature that the request body is also considered for caching. This is important for POST request caching.

msavy:
Sounds good. Are you going to attach that to this commit, or are you going to make a new one?

bekihm:
@msavy I can attach it to this PR, if you want.

EricWittmann:
Let's do a separate PR for that.  I'll merge this tomorrow, as scheduled.  Then you can follow that up with the POST improvement.

bekihm:
@EricWittmann ok, I added only a little fixup in class CachingResourcesPolicy which used the CachingPolicy.class as key name (CachingResourcesPolicy.java line 35 - 39). Now it uses its own class for for the context keys.

bekihm:
@EricWittmann  I prepared the user guide documentation for the new policy. So I can open a PR for this if the policy is merged.

EricWittmann:
@bekihm merged!  Thanks for the contribution.

PR #777, opened 2019-11-20
build: Remove maven-enforcer (already disabled)

msavy:


volkflo:
@msavy You were a little faster than me :)

PR #795, opened 2020-01-07
Fix deletion of api

volkflo:
This PR solves the problem from #790 

I removed a folder with test cases that are never used (see: https://github.com/apiman/apiman/commit/50376647e89cb85c88d29b1948dc5b1d5993e45b#comments)

Thanks to @bekihm for reviewing


volkflo:
The failing snyk test seems to be another problem.

PR #802, opened 2020-02-02
Changed url of api catalog and plugin repository to new one, apiman was returning http error 500.

tevosouza:


msavy:
Hmm, is it OK to just use those links directly now? I recall we used RawGit for two reasons (1) because it fixed the content type headers, and (2) it was suitable for [potentially] higher volumes of traffic.

Assuming the content type headers don't matter, is it OK to use GH URLs directly like this nowadays?

EricWittmann:
Good point, @msavy - It's good that @tevosouza is replacing the rawgit URLs, but I think perhaps we should use this instead:

https://www.jsdelivr.com/rawgit

What do you think, @tevosouza ?

tevosouza:
I'm working on UI to add global pagination for consumers ( orgs, apis, clients, ... ) and by a chance i identified this problem.
I have no point to add and no issues to change rawgit to jsdelivr (I didn't know jsdelivr.. ).
I will make another PR changing URLs to jsdelivr as soon as possible. 

Thanks @msavy and @EricWittmann

EricWittmann:
Thanks!  If you run into any other "rawgit" URLs we should definitely change those as well.   Thanks for the contribution. :)

PR #806, opened 2020-02-08
Added pagination and minor fixes to organization search page.

tevosouza:
I have added pagination to the organization search page and some minor adjustments on the same page (limit description, page size select, wildcard search improvement). If this way of pagination is accepted I intend to add it to another searches and lists (if it fits)

Some of my changes are based on @volkflo PR #668 (special thanks)

Description limit (complete description on tooltip)
image

PageSize Selector
image

Pagination:
image

Reviews and comments are welcome.
Thx

### DIFF
--- a/manager/ui/war/plugins/api-manager/ts/consumer/consumer-orgs.ts
+++ b/manager/ui/war/plugins/api-manager/ts/consumer/consumer-orgs.ts
@@ -8,6 +8,9 @@ module Apiman {
             if (params.q) {
                 $scope.orgName = params.q;
             }
+
+            $scope.pageSize = 10;    //default page size
+            $scope.currentPage = 1; //initial page index
             
             $scope.searchOrg = function(value) {
                 $location.search('q', value);
@@ -19,6 +22,8 @@ module Apiman {
                         var body:any = {};
                         body.filters = [];
                         body.filters.push( {"name": "name", "value": "*" + params.q + "*", "operator": "like"});
+                        body.page = 1;          //default on backend for some functions
+                        body.pageSize = 500;    //default on backend for some functions
                         var searchStr = angular.toJson(body);
                         ApimanSvcs.save({ entityType: 'search', secondaryType: 'organizations' }, searchStr, function(result) { 
                             resolve(result.beans);
--- a/manager/ui/war/plugins/api-manager/css/apiman.css
+++ b/manager/ui/war/plugins/api-manager/css/apiman.css
@@ -409,6 +409,7 @@ div.page {
 }
 
 .browse-items {
+  padding: 0px;
 }
 .browse-items .count {
   margin-bottom: 15px;
@@ -416,7 +417,7 @@ div.page {
 .browse-items .item {
   position: relative;
   padding: 10px;
-  width: 330px;
+  width: 20%;
   border: 1px solid rgb(218, 218, 218);
   float: left;
   margin-right: 15px;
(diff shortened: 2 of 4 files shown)


volkflo:
Hey @tevosouza, 
very cool that you started contributing to this project.
But the implementation is not what I was expecting.

It looks to me that you implemented the paging only within the frontend by fetching always 500 Objects on one page and then do the paging in the UI.

The backend already provides paging. See this little example:
image
image

I think you should set the `pageSize` to that value that the user can choose with the dropdown.
With the returned value `totalSize` you can calculate how many pages exist.
If the user switches the page you would send the request again.

Best wishes,
Florian

manager/ui/war/plugins/api-manager/ts/consumer/consumer-orgs.ts
```diff
@@ -19,6 +22,8 @@ module Apiman {
                         var body:any = {};
                         body.filters = [];
                         body.filters.push( {"name": "name", "value": "*" + params.q + "*", "operator": "like"});
+                        body.page = 1;          //default on backend for some functions
+                        body.pageSize = 500;    //default on backend for some functions
```
volkflo: @tevosouza As in my [comment](https://github.com/apiman/apiman/pull/806#issuecomment-584013482) here I think you should set the pageSize from the dropdown. Otherwise, you would only get 500 entries in a wildcard search :)

tevosouza: Hi,
This was really my approach, even though it was already implemented in the backend, I used 2 criteria to decide

1- In a wildcard search, in theory, the user knows what he is looking for if by chance he does a search using '*' I believe that limiting the number of records returned is not a problem.
2- Because it is a search screen, I found it less traumatic to make just one call to bring the results and page only on the front end.

I believe that because they are organizations, there is hardly a use case (in my opinion) that the user needs to do a search that returns more than 500 lines and uses all this data.

bekihm: @tevosouza Maybe you can spend just a little more effort to compute the requested page size from the backend :-)
Like @volkflo said, use the pageSize from the user and calculate in the front end the requested page number. You just need a check if the current returned result from the backend allows a click on the next/previous button :-)

for the next button:
`totalSize < pageSize * page`
or more concretely:
`totalSize < (lastPageNumber * pageSize) + currentResultCount`

bekihm: @tevosouza You are almost there :-)

tevosouza: @bekihm and @volkflo, you convinced me, I'll change my approach of pagination and make a new pull-request later, by now I'll close this pull-request.1

Thanks for your comments!!

bekihm:
@tevosouza Just as hint you can also do a force push on this PR. So then you not need to close it and open a new one again.

EricWittmann:
There's no need to do a force-push.  Just a regular push of additional commits is fine.  I'll squash during the merge so we only end up with a single commit per Pull Request. :)

So in other words, even after you've made a Pull Request, you can always modify it by just making additional changes to the branch (in this case you've used `master` but you could also have made a separate branch specific to the work being done), committing those changes as normal, and then pushing the new commit(s).

I'm not sure why many folks prefer to force-push...

bekihm:
Sorry, we use force push to clean up our history by our self, just in case it is needed :-)

PR #807, opened 2020-02-12
Allows user to pass authorization header downstream

bulivlad:
The basic-authentication-policy discards the `Authorization` header before passing the request downstream causing the target system to fail authenticating the request

This change is meant to allow the used to specify whether it needs to pass the `Authorization` header downstream to the tagert system by configuring `"forwardBasicAuthDownstream": true` property on the policy. If the property values is `false` or is not set at all the policy will strip down the `Authorization` header in order to maintain the backward compatibility

### DIFF
--- a/gateway/engine/policies/src/main/java/io/apiman/gateway/engine/policies/BasicAuthenticationPolicy.java
+++ b/gateway/engine/policies/src/main/java/io/apiman/gateway/engine/policies/BasicAuthenticationPolicy.java
@@ -33,6 +33,7 @@ import io.apiman.gateway.engine.policy.IPolicyContext;
 import io.apiman.gateway.engine.policy.PolicyContextKeys;
 
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.lang3.BooleanUtils;
 
 /**
  * An implementation of an apiman policy that supports multiple styles of authentication.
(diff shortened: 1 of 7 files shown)


EricWittmann:
Out of curiosity can you explain the use-case for this?  If the BASIC auth policy is enabled, that would mean you want the Apiman gateway to perform authentication.  If the downstream service is going to perform BASIC authentication, why **also** have the gateway do it?

EricWittmann:
I should say that this change seems reasonable and I'm happy to merge it - I'm just curious about the use-case.

volkflo:
@bulivlad May you can add a screenshot or config about how to enable and configure this? :)

msavy:
One use-case I know of is to allow a subset of downstream services to extract roles from the token in to do auditing, etc, but to use the API Gateway as the canonical 'stop-go' barrier. 

This prevents you from accidentally exposing a service (or part of a service) that does not have/enforce auth.

EricWittmann:
@msavy or @volkflo any reason to not merge?

msavy:
I don't know if I'm being unobservant, but I can't see any corresponding UI element(s) for this?

bekihm:
I agree with @msavy. I think to configure this flag via UI, we need changes in the [policy template](https://github.com/apiman/apiman/blob/master/manager/ui/war/plugins/api-manager/html/policyForms/basic-auth.include) and in the [corresponding Controller](https://github.com/apiman/apiman/blob/master/manager/ui/war/plugins/api-manager/ts/policies.ts#L382).

volkflo:
I also agree with @msavy and @bekihm. It would be nice if something like that could be activated also via the UI if you need it.

EricWittmann:
Hey @bulivlad - I know it's been a little while now, but would it be possible for you to update the manager UI to add support for this new option?

EricWittmann:
Hey what do you guys want to do about this?  Worth merging without the UI update?

tevosouza:
I believe that the functionality is incomplete without the possibility of configuring it through the user interface, therefore, I think it should not be merged and because it is a very specific use case I think it is better to wait for the rework of this pull request.

PR #808, opened 2020-02-24
Allow override of Keycloak Realm & URL via environment

dbayub:
I need to be able to easily use a different realm and URL when the Keycloak server is external. This allows the use of environment variables, making them easily overridden for a docker container, while leaving the current defaults. Also fix test that did not wait for Elasticsearch to start.

### DIFF
--- a/distro/eap7/src/main/resources/overlay/standalone/configuration/standalone-apiman.xml
+++ b/distro/eap7/src/main/resources/overlay/standalone/configuration/standalone-apiman.xml
@@ -430,8 +430,8 @@
         </subsystem>
         <subsystem xmlns="urn:jboss:domain:keycloak:1.1">
           <secure-deployment name="apiman.war">
-            <realm>${env.REALM:apiman}</realm>
-            <auth-server-url>${env.SSO_URL:/auth}</auth-server-url>
+            <realm>${env.APIMAN_AUTH_REALM:apiman}</realm>
+            <auth-server-url>${env.APIMAN_AUTH_URL:/auth}</auth-server-url>
             <resource>apiman</resource>
             <credential name="secret">5af5458f-0a96-4251-8f92-08ebcc3a8aa2</credential>
             <disable-trust-manager>true</disable-trust-manager>
@@ -440,8 +440,8 @@
             <principal-attribute>preferred_username</principal-attribute>
           </secure-deployment>
           <secure-deployment name="apimanui.war">
-            <realm>${env.REALM:apiman}</realm>
-            <auth-server-url>${env.SSO_URL:/auth}</auth-server-url>
+            <realm>${env.APIMAN_AUTH_REALM:apiman}</realm>
+            <auth-server-url>${env.APIMAN_AUTH_URL:/auth}</auth-server-url>
             <resource>apimanui</resource>
             <credential name="secret">722557fd-a725-4cc0-9dff-7d09c0c47038</credential>
             <disable-trust-manager>true</disable-trust-manager>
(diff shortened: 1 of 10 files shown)


EricWittmann:
Overall this looks good.  The only thing is probably the choice of env property name.  `REALM` is perhaps a bit too generic.  @msavy thoughts?

msavy:
Agreed, perhaps AUTH_REALM or something is better?

distro/eap7/src/main/resources/overlay/standalone/configuration/standalone-apiman.xml
```diff
@@ -429,35 +429,34 @@
           <web-context>auth</web-context>
         </subsystem>
         <subsystem xmlns="urn:jboss:domain:keycloak:1.1">
-          <realm name="apiman">
-            <realm-public-key>MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAxyG61ohrfJQKNmDA/ePZtqZVpPXjwn3k3T+iWiTvMsxW2+WlnqIEmL5qZ09DMhBH9r50WZRO2gVoCb657Er9x0vfD6GNf/47XU2y33TX8axhP+hSwkv/VViaDlu4jQrfgPWz/FXMjWIZxg1xQS+nOBF2ScCRYWNQ/ZnUNnvrq8dGC2/AlyeYcgDUOdwlJuvgkGlF0QoVPQiRPurR3RwlG+BjL8JB3hbaAZhdJqwqApmGQbcpgLj2tODnlrZnEAp5cPPU/lgqCE1OOp78BAEiE91ZLPl/+D8qDHk+Maz0Io3bkeRZMXPpvtbL3qN+3GlF8Yz264HDSsTNrH+nd19tFQIDAQAB</realm-public-key>
```
msavy: If we just change this, won't it break the OOTB quickstart experience?

EricWittmann: Good question - @dbayub is the `realm-public-key` not needed?  It's possible it isn't needed...

EricWittmann:
> 
> 
> Agreed, perhaps AUTH_REALM or something is better?

I was thinking maybe all the ENV variables should be prefixed with `APIMAN_` to avoid problems with other applications having similar needs.  Thoughts @dbayub and @msavy ?

tevosouza:
> > Agreed, perhaps AUTH_REALM or something is better?
> 
> I was thinking maybe all the ENV variables should be prefixed with `APIMAN_` to avoid problems with other applications having similar needs. Thoughts @dbayub and @msavy ?

I agree with @EricWittmann and I was thinking about standards (usually on docker) used by keycloak environment variables KEYCLOAK_USER, KEYCLOAK_PASSWORD, KEYCLOAK_IMPORT, ... all of then were prefixed with the application name.

EricWittmann:
OK @dbayub - I think this is good but with one change - please rename the ENV variables to be prefixed by `APIMAN_AUTH_` ... so I think these are the new vars:

* APIMAN_AUTH_REALM
* APIMAN_AUTH_URL

Good stuff.

EricWittmann:
This builds locally but I can't test it because of that ambiguous `RestExceptionMapper` problem.  I'll try to get that fixed so that I can verify this PR and merge it.

volkflo:
@EricWittmann I looked into this yesterday a little bit, but was not sure what is causing this exactly :/

tevosouza:
I'm not an experienced java developer (I'm a c# developer), but I tried too and got no success. 
This project makes me learn a lot of new things about java, maven and API management.

EricWittmann:
OK thanks for looking into it, both of you.  I have about 1hr right now and am going to see if I can figure it out. :)  Wish me luck!

EricWittmann:
So the refactoring of the manager api rest exceptions into their own maven module is the culprit here.  The `apiman-manager-api-rest-1.5.6-SNAPSHOT.jar` artifact, once built, contains the java classes from both the `apiman-manager-api-rest` **and** the `apiman-manager-api-rest-exceptions` modules.  It's probably the bundle plugin doing this - I'm still working on a solution.  Will report back later.

EricWittmann:
OK I've fixed the `RestExceptionMapper` problem - when I get some more time I'll try this PR again locally.

EricWittmann:
So I was able to validate this locally by doing a full build with `-Pinstall-all-wildfly10`.  I was able to authenticate as normal, but when I tried a normal (non-admin) user I was not able to create an Organization (I got a 403 error).  I suspect the problem was not caused by this PR, but I can't be sure until I do some further testing.

volkflo:
This is probably a result of my REST changes. If I remember this correctly we decided to allow creation of organizations only for admins. Otherwise a user can spam the the UI with many organizations.
But we can change this back if this behavior is not wanted :)

EricWittmann:
OK thanks @volkflo - I'll open an issue and we can discuss there.  I'll go ahead with the merge of this...

PR #809, opened 2020-02-26
Apiman security

dbayub:
Security enhancements for the user interface:

1. Do not expose the unencrypted password used to access the gateway. This does not apply to the initial setting.
2. To the extent possible, validate user input in the UI before submitting changes.
3. Use forms for all user input to support angular form validation.
4. Always build the translations javascript file from the properties file.


EricWittmann:
Ouch - the diff is rough on this due to (I'm assuming) line endings or something similar.  I don't suppose there's any way for you to fix that up to make the review a bit easier?

volkflo:
@EricWittmann @dbayub We worked on a refactoring of the REST API. Also for security reasons (leak of sensitive data). I will create a PR, after @bekihm reviewed it. I mentioned this now because I also did changes on the GatewayResourceImpl.java :)

dbayub:
Updated line endings. There are a mix of *nix and DOS.

EricWittmann:
@dbayub Thanks for fixing the line endings problem - the diff is much better now!  Appreciate that.  However another PR has caused some conflicts in this.  Apologies for that.  Can you rebase and fix the conflicts?  

@volkflo Can you give this the once-over and offer your thoughts?

dbayub:
Done.

volkflo:
I checked the travis logs there are many tests now crashing because of the encrpytion of the password like:
```json 
Failed tests: 
  ReregisterClientTest.Create Gateway configuration
Expected: {"endpoint":"http://localhost:7070/mock-gateway","username":"admin","password":"admin"}
     got: {"endpoint":"http://localhost:7070/mock-gateway","username":"admin","password":"$CRYPT::j8rdW76Z5gUI0I+9c8/GrA=="}
````

**Comment by EricWittmann** [comment]
> But I think it would be better to rename the env values because they are a bit generic.
> I thought I read about this in other comments before, but I can't find them at the moment.
> 
> ```shell
> env.SSO_URL
> env.REALM 
> ```

Yes - I think all new ENV variables should be prefixed by `APIMAN_`.

> With the encryption of the gateway password, I am not sure if this could cause problems with the "Export/Import" methods? @EricWittmann Any thought about this? :)

Good catch - this is not clear to me.  It's possible that encryption/decryption will need to be hooked into the import/export process.  There may be a precedent for that already there - you'll need to check on that.

**Comment by dbayub** [comment]
Those Keycloak changes shouldn't be in this change. They are in another. FYI, Keycloak recommends that the realm-public-key not be specified, the interface pulls it in from the server, and not having it allows the server to rotate keys. I had fixed the tests, but lost my change at some point. Hopefully added back now.

**Comment by dbayub** [comment]
Oh, and the changes in the UI vs the REST APIs, the pentest done was for the UI, so I had to fix that. And the package.json was changed, manager/ui/war/properties.json.

**Comment by msavy** [comment]
At the time we implemented this, the feature for getting the domain's key via API wasn't available in Keycloak, I believe -- I certainly don't object to the idea of referencing the SSO by default and pulling the domain key(s) from there.

The complication would be if Apiman can't reach the Keycloak server, but the browser redirect flow works OK,  which is a problem we occasionally see.

I definitely think defaulting to using KC to grab the key would be a much better default. We could then look at randomising the domain key for the quickstarts, for example.

**Comment by volkflo** [comment]
> Oh, and the changes in the UI vs the REST APIs, the pentest done was for the UI, so I had to fix that. And the package.json was changed, manager/ui/war/properties.json.

Sorry I missed that change in the `package.json`.
I think the REST checks can be added in the future as well :)

**Comment by volkflo** [comment]
But there are still some tests crashing

**Comment by EricWittmann** [comment]
@dbayub can you have a look at the test failures?

**Comment by EricWittmann** [comment]
Thanks for fixing the tests, @dbayub - I'll do a local build to make sure everything is good.

**Comment by EricWittmann** [comment]
@volkflo and @msavy Any last thoughts before I merge this?

**Comment by volkflo** [comment]
@EricWittmann Are the env variables now renamed? :) I think I missed a bit of the last activity here.

File: manager/ui/war/wildfly8/src/main/webapp/WEB-INF/web.xml [hunk 1]
```diff
@@ -86,10 +86,10 @@
   </security-constraint>
   <login-config>
     <auth-method>BASIC</auth-method>
-    <realm-name>apiman</realm-name>
+    <realm-name>${env.REALM:apiman}</realm-name>
```
EricWittmann: @dbayub  I think you want to remove this change for this PR, right?

EricWittmann:
> 
> 
> @EricWittmann Are the env variables now renamed? :) I think I missed a bit of the last activity here.

The ENV var changes are in another PR:  https://github.com/apiman/apiman/pull/808

But you are right - one of those changes snuck into this PR, so good catch. :)  I've commented on that part of the PR @dbayub - one of the `web.xml` changes is still here.

EricWittmann:
I resolved the conflicts and I tested locally - everything checks out (smoke tests).  Thanks @dbayub for this contribution.

Issue #803, opened 2020-02-05
Minor UI improvement


tevosouza:
Hello guys!
What do you think about this kind of improvement? 
I'm developing pagination on the UI side for some lists that were returned by APIs.
![image](https://user-images.githubusercontent.com/58958299/73889120-aff56c80-484d-11ea-9468-e96538bc4cb6.png)

volkflo:
Hey :)

this looks good, a while ago I used another workaround to get all Orgs/APIs not only the first 20.
I did a wildcard search and set the page size to 10000 if I remember that correctly.

See https://github.com/apiman/apiman/pull/668

Maybe this helps you, but I think your approach is the better one :)

bekihm:
Looks like a very good improvement 👍

tevosouza:
Hi!
> Looks like a very good improvement  I'm not sure if we need also an implementation for the backend side for huge data.
> Hey :)
> this looks good, a while ago I used another workaround to get all Orgs/APIs not only the first 20.
> I did a wildcard search and set the page size to 10000 if I remember that correctly.
> See #668
> Maybe this helps you, but I think your approach is the better one :)

@volkflo
I saw your PR and it was the trigger for my "pagination" idea..thanks for it!

> Looks like a very good improvement  I'm not sure if we need also an implementation for the backend side for huge data.
@bekihm, I agree with you, I think we no need to spend effort on it right now, maybe another time.

Well, I'll improve my code and make a new PR soon

bekihm:
@tevosouza 
Here is the corresponding [Java 11 issue](https://github.com/apiman/apiman/issues/796).
@msavy has said there he is working on this but maybe you can report your problems there.

tevosouza:
> @tevosouza
> Here is the corresponding [Java 11 issue](https://github.com/apiman/apiman/issues/796).
> @msavy has said there he is working on this but maybe you can report your problems there.

haha!
I'm "just" trying it and discovering how difficult it is. This is a job for apiman gurus .. LoL

msavy:
Still working on it (Java 11), but due to personal life commitments (family
illness etc) I've not been able to spare much mental energy for Apiman.

I'll get back on it soon.

tevosouza:
Hi @volkflo and @bekihm 
The first part of the 'pagination-project' is done!
I've made a pull request #806 with some adjustments, pagination and page size selection only on organization search, if the way that I've used is accepted I'll implement it on other pages and lists (if it fits).

[...]

PR #810, opened 2020-02-29
Added orgs pagination to search

tevosouza:
I have added pagination to the organization search page and some minor adjustments on the same page (limit description, page size select) 

Some of my changes are based on @volkflo PR #668 and discussion in PR #803 

Description limit (complete description on tooltip)
image

PageSize Selector
image

Pagination:
image


manager/ui/war/plugins/api-manager/html/consumer/consumer-orgs.html
```diff
@@ -48,14 +48,24 @@ <h1 class="consumer-top-header" apiman-i18n-key="find-org">Find an Organization<
...
+              <option value="25">25</option>
+              <option value="50">50</option>
+            </select>
+          </div>
+
+        </div>
 	        <div class="item" ng-repeat="org in orgs">
 	          <div class="title"><i class="fa fa-shield icon"></i><a href="{{ pluginName }}/browse/orgs/{{ org.id }}">{{ org.name }}</a></div>
-	          <div class="description" title="{{ org.description }}">{{ org.description }}</div>
+	          <div class="description" title="{{ org.description }}">{{ org.description | limitTo: 60 }}<span ng-show="org.description.length > 60">...</span></div>
```
EricWittmann: I'm fine with this approach.  I just wanted to point out that it may also be possible to accomplish this using CSS (as an alternative approach).  There are `overflow` and `text-overflow` CSS properties for this.  They are a little bit tricky, however, so not sure about this specific use-case.  Not something I was aware of when I originally wrote this page. :)

EricWittmann:
This looks pretty reasonable to me.  I'll give this a couple of days for review from others before merging.

Thanks, @tevosouza !

manager/ui/war/plugins/api-manager/html/consumer/consumer-orgs.html
tevosouza: thanks for the hint, I've learned a new thing today! ;)

volkflo:
For me, this looks good now :)
It would be cool to have this then also for the page that shows all APIs.

EricWittmann:
Merged.   Thanks @tevosouza for the contribution.

PR #811, opened 2020-03-04
Refactoring REST API

volkflo:
Hi together, 

as I mentioned in other issues here this is the PR for the REST API.
We did a refactoring to improve the security and make the REST API less chatty for users without the correct permissions.

This fixes:

- Leak of API-Keys via /export and /download for non-admin users
- Leak of sensitive data (e.g. unmanaged endpoints)
- Leak of user data (e.g. email)

On top, we integrated the Swagger of the REST API in the UI.
Also, we updated the SwaggerUI Version from 2 to 3.
image
image

The upgrade of the SwaggerUI was made by us a long time ago and I squashed all our commits about this into one.

So let me know what you are thinking about all of this :)

Special thanks @bekihm for reviewing all of this :)


manager/api/config/src/main/java/io/apiman/manager/api/config/Version.java
```diff
@@ -42,32 +41,29 @@
      */
     public Version() {
     }
-    
-    @PostConstruct
-    public void postConstruct() {
-        load();
-    }
```
EricWittmann: What prompted this change?

manager/api/rest-impl/src/main/java/io/apiman/manager/api/rest/impl/ActionResourceImpl.java
```diff
@@ -89,17 +77,19 @@
 
     @Inject @ApimanLogger(ActionResourceImpl.class) IApimanLogger log;
 
+    private OrganizationResourceImpl organizationResource;
+
```
EricWittmann: Is this used?

manager/api/rest-impl/src/main/java/io/apiman/manager/api/rest/impl/PermissionsResourceImpl.java
```diff
@@ -1,99 +0,0 @@
...
-import javax.enterprise.context.ApplicationScoped;
-import javax.inject.Inject;
-
-/**
- * Implementation of the Permissions API.
- * 
- * @author eric.wittmann@redhat.com
- */
-@ApplicationScoped
-public class PermissionsResourceImpl implements IPermissionsResource {
```
EricWittmann: Sorry but I missed how this can be removed... (this diff is a bit large) :)  Is this API no longer needed, or have these methods been moved to another resource impl?

manager/api/config/src/main/java/io/apiman/manager/api/config/Version.java
volkflo: The problem was in `manager/api/war/tomcat8/src/main/java/io/apiman/manager/api/war/tomcat8/Tomcat8ApiManagerApplication.java` in this line:
```java
 beanConfig.setVersion(new Version().getVersionString());
```
The result of `new Version().getVersionString()` was always null because the postConstruct was somehow not called.
And so Swagger complained about the missing version number in the JSON/YAML.

I removed this postConstruct section and moved the call of `load()` into the two getters:

```java
 /**
     * @return the versionString
     */
    public String getVersionString() {
        load();
        return versionString;
    }

    /**
     * @return the versionDate
     */
    public String getVersionDate() {
        load();
        return versionDate;
    }
```
So now we are sure that this is called every time if `versionString` or `versionDate` are null :)

manager/api/rest-impl/src/main/java/io/apiman/manager/api/rest/impl/ActionResourceImpl.java
volkflo: No, I don't think so. Well paid attention 👍 

I will check this again and provide a fixup soon :)

manager/api/rest-impl/src/main/java/io/apiman/manager/api/rest/impl/PermissionsResourceImpl.java
volkflo: No problem, I know this is a big PR, maybe I had to split it up into separate ones :)

I removed all operations for the current user because they are mostly duplicate resources like:

* http://localhost:8080/apiman/permissions/ 
  * This endpoint returns all of the permissions assigned to the currently authenticated user.
* http://localhost:8080/apiman/permissions/userId/
  * This endpoint returns all of the permissions assigned to a specific user.

So I also removed the /currentuser endpoint.
Then I changed the order a little bit. The permission are now in the user api directly.

* http://localhost:8080/apiman/users/userId/permissions

manager/api/config/src/main/java/io/apiman/manager/api/config/Version.java
EricWittmann: OK that's fine.  I probably would have preferred a change to `Tomcat8ApiManagerApplication` instead:

```java
Version version = new Version();
version.postConstruct();
beanConfig.setVersion(version.getVersionString());
```

But honestly it's not a huge deal.

manager/api/rest-impl/src/main/java/io/apiman/manager/api/rest/impl/PermissionsResourceImpl.java
EricWittmann: OK thank you I understand.  So just to be clear, the *logical* consequence of this change  (from an API perspective) is that "currentuser" is now sort of a reserved userId.  Not really, but in the framework of the REST API it is.  So for example, if my username is "ewittmann" then these are all valid REST endpoints:

`/users/ewittmann`
`/users/ewittmann/organizations`
`/users/ewittmann/apis`
`/users/currentuser/info`

And then all the other currentuser stuff from before is gone and instead you have to route the requests for that stuff using the currently authenticated user's actual **userId**.

I'm cool with all of that except the new `/users/currentuser/info` endpoint, since that's a little bit weird.  Not weird enough for me to not merge the changes, but still - just a little weird. :)

On balance, I think the changes to the API are very good - and I'll live with the tiny bit of weirdness!

EricWittmann:
Sorry but another PR merge caused a conflict in `apiman.css` - hopefully you can resolve that without too much trouble.

manager/api/config/src/main/java/io/apiman/manager/api/config/Version.java
volkflo: Just out of interest why would you prefer the other changes? :)

manager/api/rest-impl/src/main/java/io/apiman/manager/api/rest/impl/PermissionsResourceImpl.java
volkflo: Yes, all of these are valid endpoints.
Everything of the currentuser/* is moved to /user/{userId}/*.
But only you are allowed to access your stuff.

So if I would call `/users/ewittmann/organizations` this will lead to a forbidden status code. Only an admin is allowed to view all other users stuff.

And yeah `/users/currentuser/info` is a workaround but we have had no better idea at this moment. 

I try to explain it and maybe you have a better idea :)

The problem with these changes is that the UI has to know the username for most of the requests now. Like the whole "My Stuff" Pages (/users/fvolk/apis).

But the UI used `currentuser/info` to get the username to assign the current user to the pages.
So we moved this to `users/currentuser/info` so the UI still gets a username and then we can put this username in all new requests. Otherwise, we are not able to find out the username without more changes in the UI.

EricWittmann: Yeah - I totally understand everything you've explained (including the workaround).  And I also have no better suggestion.  :)  So let's stick with what you have.

manager/api/config/src/main/java/io/apiman/manager/api/config/Version.java
EricWittmann: It was nice having the `Version` as a CDI bean.  I'm not a huge fan of lazy loading as part of the getters.  It's error prone (e.g. if we added more properties to the class we need to remember to call `load()` in every getter) and typically not best-practice IMO.  Sometimes it's necessary.

Another valid approach would be to make `Version` a singleton.

But again, I don't feel strongly enough about it to ask you to change it. :)

volkflo: Okay, that sounds good :) 
But it is good to know what was your idea about this. Maybe we can refactor this one time in the future, but unfortunately, our resources are also limited.

volkflo:
So I rebased all of this on the current master and merged some CSS. Hopefully, this will not break the whole design 🙏

PR #815, opened 2020-03-07
Minor fixes

tevosouza:


tevosouza:
> LGTM
> I wondering only about the console.log statements which were there before.
> See [manager/ui/war/plugins/api-manager/ts/plan/plan.ts](https://github.com/apiman/apiman/pull/815/commits/323d7a7f8e42a2ce075f15449b279294863c7044#diff-8a733750778264967d40652e067c7307L142)

Hi @bekihm 
I removed this log.console because when I read the code it only exists in the plan controller.
maybe for debugging?

manager/ui/war/plugins/api-manager/ts/modals.ts
```diff
@@ -59,7 +59,8 @@ module ApimanModals {
         function ($location, $rootScope, $scope, $uibModalInstance, OrgSvcs, Configuration, PageLifecycle, client, params) {
         
         $scope.confirmClientName = '';
-        
+        $scope.client = client;
```
volkflo: Just for understanding what does these line do:

```javascript
$scope.client = client;
$scope.org = org
$scope.plan = plan;
```

tools/server-all/pom.xml
```diff
@@ -240,14 +240,13 @@
                   <tasks>
                     <antversion property="antversion" />
                     <echo>Ant Version: ${antversion}</echo>
-
                     <property name="appserver.id" value="tomcat8" />
                     <property name="apiman.tomcat.download.url"
-                      value="http://apache.cs.utah.edu/tomcat/tomcat-8/v8.0.33/bin/apache-tomcat-8.0.33.zip" />
+                      value="http://apache.cs.utah.edu/tomcat/tomcat-8/v8.5.51/bin/apache-tomcat-8.5.51.zip" />
```
volkflo: I am not sure what these `server-tools` exactly do, but have you tried if we are getting problems if we would upgrade to tomcat 9 or even 10?

manager/ui/war/plugins/api-manager/ts/modals.ts
tevosouza: Hi @volkflo 
I've added this because of it not accessible from a template with {{ client.name }} or {{ org.name }} in placeholders of deletion modal.

When modal was show the textbox placeholder was empty. These lines are the fix ;)

tools/server-all/pom.xml
tevosouza: I didn't try to upgrade the version of tomcat, I only fixed it to optimize and make easer my tests, but I'll try it in a few days.

PR #816, opened 2020-03-10
ci(travis): Use double quotes on travis actions

msavy:
The action '-Dapiman-test.type=es -Dapiman.gateway-test.config=servlet-es' was failing because it is now being interpreted as a single string seemingly. 

Trying double quotes to see whether that works.


msavy:
I believe this has worked

Issue #817, opened 2020-03-11
Non-admin users cannot create organizations


EricWittmann:
In testing a recent PR I noticed that I wasn't able to create an organization unless I was logged in as an admin.  @volkflo suggested that was due to a recent change that his team made for that specific purpose.  We should discuss.

The UI assumes that all users can create organizations.  I think what we should do is make this configurable (globally).  By default I think the old behavior should be restored (all users can create Orgs) but with a configuration option the new behavior (admins only) could be enabled.  Along with that change we would need to update the UI so that the option to create an Org only exists (e.g. on the Dashboard) when it's allowed.

volkflo:
Yes, I checked this again. See here: https://github.com/apiman/apiman/blob/e14d18ec74b20a675591fa0384ead48e55519f3c/manager/api/rest-impl/src/main/java/io/apiman/manager/api/rest/impl/OrganizationResourceImpl.java#L139-L141

In our UI is this the case, the create button is hidden if you are not logged in as admin.
But I agree with your suggestion, that It would be cool the make this configurable somewhere (not sure where?).

EricWittmann:
Yeah I can't remember how configuration options work in Apiman or even if we had something like that.  I'll need to dig in a bit to refresh my memory. :)

tevosouza:
If I understood correctly, organizations are the "base" for all other things in apiman. I'm not sure if deny the creation of the organization to the user even if they don't have admin privileges is a good idea.
I'm assuming that if the user is logged on the system is for doing some action and the access was granted by one administrator and if the organization is the "base" the user could be blocked to use the system (can't do any other action).

EricWittmann:
Yeah that was the idea for Orgs for sure (basically the same concept as GitHub).  But I also understand that some companies may want to limit the creation of Orgs to only admins.  Then those admins will be responsible for adding other users to those orgs.  This would centralize the admin and configuration of organizations.  I can see that as a valid use-case.  So I'm fine with the option, but it shouldn't be the default IMO.

volkflo:
@EricWittmann I think we can close this now? :)

PR #819, opened 2020-03-12
Config option to enable admin-only creates for organizations

EricWittmann:


EricWittmann:
This attempts to address https://github.com/apiman/apiman/issues/817

EricWittmann:
@volkflo and @tevosouza - how does this look to you?

volkflo:
@EricWittmann Cool that you have already implemented this 👍  For me this is fine :)

There are just two little things:
I stumbled over this for the UI.
https://github.com/apiman/apiman/blob/e14d18ec74b20a675591fa0384ead48e55519f3c/manager/ui/war/plugins/api-manager/ts/dash.ts#L4-L11
I am not sure if something of this is still used somewhere.

And the other point is that it would be great if you could add the new system property into the `apiman.properties` files directly. So we know in a few weeks that there is an option we can configure without looking into the code :)

EricWittmann:
Yeah I didn't hook the option into the UI yet.  I'll need to do that as well - thanks for tracking down where that would be done!  I honestly can't remember why Org Creation was disabled for those platforms.... technical debt!

tevosouza:
@EricWittmann this is fine for me!
Thanks a lot!

EricWittmann:
OK @tevosouza and @volkflo - please have one last look at this PR.  I updated it include UI changes.  Opinions welcome. :)

tevosouza:
@EricWittmann looks good to me, I have no comments to add
Thanks!

EricWittmann:
Good point @volkflo - I've reverted `yarn.lock` since I didn't make any `package.json` changes.  I'm going to merge this and then add the property to `apiman.properties` as you suggested in a separate PR (along with a fix for the Swagger UI problem when running in Wildfly).

PR #825, opened 2020-03-20
Gulp build fix and some i18n adjustments

tevosouza:


volkflo:
@tevosouza Thank you for the fix, the menu entry was just a copy & paste work :)

PR #831, opened 2020-04-03
Fixed the test of gateway that broke travisCI build

tevosouza:


volkflo:
@tevosouza and @EricWittmann 
At the moment I don't see any travis tests running here:

PR #832, opened 2020-04-04
Reverted #809

EricWittmann:
OK @volkflo and @bekihm  and @tevosouza - have a look.  This is a reversion of #809.  Can you verify?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,3 @@ gwt-unitCache
 npm-debug.log
 .metadata/*
 *.log
-manager/ui/war/apiman/translations.js
--- a/manager/api/gateway/src/main/java/io/apiman/manager/api/gateway/rest/RestGatewayLink.java
+++ b/manager/api/gateway/src/main/java/io/apiman/manager/api/gateway/rest/RestGatewayLink.java
@@ -15,6 +15,7 @@
  */
 package io.apiman.manager.api.gateway.rest;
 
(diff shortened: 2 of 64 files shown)


volkflo:
I think this is good for the moment and makes the UI usable again :)
I hope this gets reworked by the author because I think these are nice changes if they are checked again. But you have to compare what validation was already done

bekihm:
Thanks @EricWittmann for the revert 👍 
I think we will test tomorrow :-)

Maybe we have a chance that the author reworks the policy validation again see [here](https://github.com/apiman/apiman/pull/833)

tevosouza:
I've made superficial tests and I think that is ok, I have found one problem with API definition (wsdl), but it's not related to PR #809.
Thanks @EricWittmann for the revert ;)

volkflo:
@tevosouza If the WSDL Problem is that you cannot delete it correctly then I will provide a PR soon :)

EricWittmann:
OK I wasn't sure if this was the correct PR to revert, because I couldn't find any regexp changes in there.  There are some validation changes for sure, but also a lot of security related fixes.  So please give this some testing and make sure it's what needs to be reverted.  :)

EricWittmann:
Oh, and just "@EricWittmann" here when we're ready to merge!

volkflo:
@EricWittmann I tested this, everything in the UI is working again 👍

PR #838, opened 2020-04-07
Updated the swagger WSDL code to be namespace aware

EricWittmann:
This update should make the algorithm much more robust when processing WSDL files.


EricWittmann:
@volkflo @tevosouza @bekihm Have a look at this update and let me know what you think.  It should do the same thing as before but better. :)

EricWittmann:
Note: there are some failing unit tests, but I think failing due to another change, not mine.  I'll fix those and update this PR.

EricWittmann:
OK I was missing a couple things in this PR - will work on it tomorrow and update here when ready. :)  Stay tuned.

EricWittmann:
OK this is now ready to review.  I believe all unit tests are green and the code actually makes sense.  I had to dig out all of my old WSDL/XML/SOAP knowledge from years ago, dust it off, shake it out, and try to make sense of it.  :)  But I think this is now the right approach!

tevosouza:
LGTM!
Thanks @EricWittmann

EricWittmann:
Thanks everyone.  What a great throwback problem!  :)

PR #866, opened 2020-07-13
Sidebar Navigation

tevosouza:
Added a fixed (very simple) sidebar navigation to improve UX
I've removed some tabs of administration pages (redundancy) and adjusted some alignments.
Change proposes are welcome ;)

Preview:
image

### DIFF
--- a/manager/ui/war/plugins/api-manager/html/sidebar/sidebar.include
+++ b/manager/ui/war/plugins/api-manager/html/sidebar/sidebar.include
@@ -1,0 +1,71 @@
+<div class="sidebar" ng-controller="Apiman.SidebarController">
+
+<div class="sidebar-category">Menu</div>
+
+ <a href="{{ pluginName }}">
+    <i class="icon-text fa fa-lg fa-tachometer"></i>
+    <span class="icon-text" apiman-i18n-key="dashboard">
+        Dashboard
+    </span>
+</a>
+  
+  <a href="{{ pluginName }}/browse/orgs">
+    <i class="icon-text fa fa-lg fa-shield"></i>
+    <span class="icon-text" apiman-i18n-key="organizations">
+        Organizations
+    </span>
+</a>
+
+ <a href="{{ pluginName }}/users/{{ currentUser.username }}/clients">
+    <i class="icon-text fa fa-lg fa-gears"></i>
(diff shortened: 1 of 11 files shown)


volkflo:
This looks like a nice change :)

Our sidebar is a little more complex and we created it for our needs, so I will not contribute this.
But I added the HTML and ts/js in a [github gist.](https://gist.github.com/volkflo/bfdec58c583a95cf8097878579aec911) Maybe someone will find it useful.

tevosouza:
Very good! ;)
I'll take a deep look later, COVID-19 pandemic is killing all my time due to several "controls" that I working to implement in record time on my employer.

Thank you for sharing your code =)

tevosouza:
Please, don't merge yet I've found one bug when scrolling, and I'm reworking it.

volkflo:
I did a quick test and everything is working as expected.
image
The problem for me is that everything is aligned to the top left side. IMO this looks not so nice as centered, but that's a matter of personal preference :)

msavy:
I think this is pretty cool!

Thoughts, @EricWittmann 

I assume the admin segments hide when you are logged in as non-admin?

volkflo:
Ahh good point.

@tevosouza You can copy the admin-hide from my gist if you haven't done it already :)

manager/ui/war/plugins/api-manager/html/sidebar/sidebar.include
```diff
@@ -0,0 +1,71 @@
...
+ </a>
+
+ <a href="{{ pluginName }}/users/{{ currentUser.username }}/apis">
+    <i class="icon-text fa fa-lg fa-puzzle-piece"></i>
+    <span class="icon-text" apiman-i18n-key="apis">
+        APIs
+    </span>
+ </a>
+
+ <div id="sys-admin-area" ng-show="isAdmin">
```
tevosouza, line 33: I've already did it here! ;D

EricWittmann:
This looks pretty good to me.  If it were me I'd probably play with the colors and fonts a little bit to better match the masthead, but if everyone else is happy with this then I am too.  :)

Sorry for the late response - I've been out on vacation for a few weeks.  :->

If there are no objections this week I'll merge this on Friday.

PR #889, opened 2020-08-13
Replaced wildfly 11 with wildfy 20 and removed elasticsearch from quickstart

EricWittmann:
This pull request does two main things:

* Upgrade from Wildfly 11 to Wildfly 20
* Remove Elasticsearch completely from the quickstart

After we merge this, the following tasks need to be done:

1. Release a new version of Apiman
2. Update the web site to remove the Wildfly 10 and 11 instructions and replace with Wildfly 20 instructions
3. Create, build and push new `apiman/on-wildfly` docker image
4. Discuss/vote on removing support for EAP7 and Wildfly 10


EricWittmann:
@volkflo @bekihm ^^^

EricWittmann:
I could use some sanity checking on this.  I tried it locally using the (updated) `-Pinstall-all-wildfy20` build profile and everything seems to work as expected.  But if one or more of you fine people could give it a try I would appreciate it!

volkflo:
Wow, that was fast :O
I didn't expect that you get this all done in just one day.

Just one idea: Wouldn't it be easier for the future if we name it just `wildfly` instead of `wildfly20`? 
If we want to update wildfly in the future we to rename all the files again.

I will try to checkout this and test it locally asap :)

bekihm:
@EricWittmann Thanks for the great work 👍 I tried out the setup, it works from my perspective.

About the code:
- We have to modify the project readme to match the wildfly20 setup
- I found out that io.apiman.distro.es package is also used from the apiman-distro-tomcat8 package. So we cannot delete the package so easy. I think we have two possibilities we rework the [setup page of apiman](http://www.apiman.io/latest/download.html) and delete the apiman-distro-tomcat8 feature. Or we have to modify the apiman-distro-tomcat8 so that it uses an external elasticsearch for which we have to provide the manual on the [apiman page](http://www.apiman.io/latest/download.html).

volkflo:
In my opinion, one quickstart setup is enough to try out apiman :)
If the user then wants to move to a production setup I think it is okay that there is no quickstart for tomcat or other platforms.
Using e.g. tomcat would then require a full es setup or RDBS.

EricWittmann:
Thanks for the feedback.  I'm inclined to really trim down the # of distributions to just these:

1. Wildfly (Latest)
2. Wildfly + Vert.x Gateway

We can remove the others which solves the ES problem and also makes it easier to maintain things going forward.  

I'd also like to remove the EAP and Wildfly10 modules if there are no objections.  @msavy what do you think?

msavy:
That seems reasonable to me. The only question in my mind is whether we should preserve EAP also. 

Let's leave it open for a week or two and see whether anyone objects?

distro/wildfly20/src/main/resources/overlay/standalone/data/h2/keycloak.trace.db
```diff
@@ -0,0 +1,75 @@
+2020-08-13 11:53:33 jdbc[3]: exception
```
msavy, line 1: You should likely remove this

msavy: Don't think you need keycloak.mv.db either

msavy: Hmm, might be wrong on that latter one? Double check 😂

pcornish:
Hi @EricWittmann,

I'd like to voice support for retaining the Tomcat distro. We have a number of folks that have standardised on their app server and Tomcat is still a popular choice.

I completely understand the motivation for reducing the maintenance overhead by having fewer distros to maintain, though. If it does turn out that the project is going to remove the tomcat8 distro then I think it's right that we provide a migration path for folks who are on Tomcat to the supported distros. In particular we should think about writing some technical guidance for folks for whom Wildfly is not a technology with which they are familiar. Some of these elements will be nontrivial for folks if we require them to make the transition - examples to think about would be how to replicate operational configuration for their gateway from Tomcat, e.g. as access logging, performance/thread tuning, module/classpath knowledge etc. - I know all of this sounds like it's a Wildfly not an apiman concern, but the reality is that the project's choice of app server does require that apiman users know this stuff to actually use the product in production.

If we are going to drop Tomcat, could we look at this as a breaking change and bump the major version to 2.0. This, per semver, would at least indicate the size of the change to existing users.

Happy to discuss further.

Pete

volkflo:
Hi @pcornish 

we are also using tomcat in production and not wildfly :)
The only idea was to drop support for the quickstart tutorial with tomcat + embedded elasticsearch.
So if you want to try apiman in the future you have to use wildlfy.

Tomcat with a full elasticsearch or RDBS setup will be still supported.

Or do you need the quickstart with tomcat?

Hope this makes it a bit clearer.

pcornish:
😌 ah, misunderstood! 

/ end impassioned Tomcat speech

No, I don't think Tomcat needs to be part of the quickstart, but perhaps we could add an `apiman-quickstarts` repo and allow contributions there? Doesn't mean they have to end up on the official site. Perhaps `apiman-community-quickstarts` if that makes the implied 'support' more clear?

volkflo:
Not sure if we need community quickstart projects. I think for the already few contributors it easier to focus on one quickstart and the existing production setups :)

distro/wildfly20/src/main/resources/overlay/standalone/data/h2/keycloak.trace.db
EricWittmann: I probably don't need the .trace.db file - but the .mv.db file is the keycloak database with all of the proper settings and everything already configured.  So that one is definitely needed.  H2 has moved from *.h2.db files to *.mv.db files.  Some sort of new table or file format I guess.

msavy:
Agreed. Has my thumbs up unless we get objections.

EricWittmann:
I like the idea of a community quickstarts repository, even though I don't expect a bunch of contributions (would be happy to be wrong).  But it would provide a place to move the existing quickstarts out of the main repo without losing them entirely.

EricWittmann:
Also a note to @pcornish - the next release will, indeed, be 2.0.0 due to some significant Elasticsearch changes that @volkflo and @bekihm have made (upgrading to a newer version of ES).

@msavy I think we can move the EAP and older Wildfly quickstart distributions into the proposed community quickstarts repo without waiting for additional community feedback.  wdyt?

msavy:
Just for quick start seems ok to me

PR #892, opened 2020-08-21
Moved some distros and renamed Wildfly20

EricWittmann:


volkflo:
@EricWittmann Now I am a bit confused :D 

It looks like you moved more than the quickstart now. It seems that you moved the "distro" itself.
I try to explain what I mean:

The problem was only the apiman-es distro which is not compatible with the new ES 7 changes.
Now you moved the whole e.g. tomcat package out of the main repository.
So the name apiman-community-quickstarts is a bit misleading because it also affects the production setup files for e.g. tomcat.

So my idea would be to move tomcat back into the main apiman repo (because it is used in production from @pcornish and us) and maybe rename the new apiman-community-quickstarts into apiman-community-distros?

EricWittmann:
In my mind the "distro" is synonymous with "quickstart".  When looking at the Tomcat8 distro, it contains configuration files that I associate with the building of a downloadable Tomcat Quickstart - not necessarily a production deployment.  However, if you are relying on that distro when building for production we can just add it back in.  No worries.

The actual WAR files that are packaged into the various quickstarts have not been removed - that's the part I thought you were likely using in your prod builds.

Note that for most of the other distros (e.g. Wildfly) we include lots of things that are very much quickstart-only, like the pre-built Keycloak database and the H2 based Apiman datasource config file.

In any case, I'm happy to restore the Tomcat distro - perhaps submit a PR with exactly what you'd want it to be (which is perhaps just whatever it was before).  :)

PR #894, opened 2020-08-26
Update readme and fix scripts 

volkflo:
I changed a little bit on how wildfly is unzipped and copied so that the scripts work if we update wildfly again.
Additionally fixed a little ES bug we accidentally committed.


EricWittmann:
Good changes!

PR #908, opened 2020-09-17
Fix #907 - broken plugin page

### LINKED ISSUE #907: Apiman 2.0.0.Final pointing to old plugin versions
Does anyone remember how this works?  When I install Apiman 2.0.0.Final and go to the Plugins admin page to install a plugin, the plugin versions are all 1.5.8-SNAPSHOT rather than 2.0.0.Final.  I can't remember where that happens.  There's a registry of them somewhere.  I haven't looked into it yet...

@apiman/maintainers

volkflo:
Hi @EricWittmann,

I looked into the plugins page and reverted a commit from Marc from 2018.
This fixes the routing but breaks the design (two buttons instead of nice tabs).
I think the design is not that important for the moment.

The only problem now is that the 2.0.0 Plugins can not be found.
But I am not sure if this was caused by the release pipeline (they are uploaded to maven central?)?

image

### DIFF
--- a/manager/ui/war/plugins/api-manager/html/admin/admin-export.html
+++ b/manager/ui/war/plugins/api-manager/html/admin/admin-export.html
@@ -10,12 +10,6 @@
         </div>
       </div>
       <div class="row">
-       <!-- Side Navigation -->
-       <!-- 
-         <div class="col-md-3">
-          <div ng-include="'plugins/api-manager/html/admin/admin_tabs.include'"></div>
-        </div>
-        -->
         <!-- Admin Content -->
         <div class="col-md-9 admin-content">
           <div class="container-fluid">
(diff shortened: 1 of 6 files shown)


EricWittmann:
Looking at your PR, I bet the relevant change here is the `href` -> `ng-href`.  I may try to make ONLY that change and see if that fixes the problem.

EricWittmann:
As for the loading of the plugins - you are right, they are not in maven central!  I was certain I did that, but they are clearly not there, so I'll look into that again.

volkflo:
Yes, we thought the same with the ng-href. I am on it :)

EricWittmann:
Plugins should be in Maven Central now.

volkflo:
So, I found a simpler fix for this :)
And 2.0.0 plugins are also here: 
image

EricWittmann (approved):
LGTM - ready to go?

volkflo:
Done :) Seems the permissions are also good now :)

PR #910, opened 2020-09-18
PollingEsRegistry overwrites EsRegistry indices on index creation

bekihm:
We had found an issue with the `PollingEsRegistry` which overwrites the index creation of the `EsRegistry`. So indices are missing if the `PollingEsRegistry` is used instead of the `EsRegistry`


volkflo (approved):

EricWittmann (commented):
lgtm

PR #914, opened 2020-10-09
Add backend for the new developer portal

volkflo:
Here is the backend for the developer portal.
At the moment only implemented for elasticsearch as storage.


README.md
```diff
@@ -115,3 +115,15 @@ tools/src/eclipse/apiman-eclipse-formatter.xml
 
 You should be able to import that guy straight into Eclipse by going to
 *Window->Preferences :: Java/Code Style/Formatter*
+
+## Developer Portal
+
+Apiman also comes with an developer portal. There you can allow Developers to access your APIs. Developers can view and test your APIs to develope their own Apps.\
```
EricWittmann, line 121: couple typos:  should be "...with a developer portal..." and "...APIs to develop..."

volkflo: Fixed. Thanks for the hint :)

EricWittmann (approved):

EricWittmann:
Looks good.  We should consider how to include the dev portal in the quickstart...

EricWittmann:
From what I can tell it's just a plain Angular app that uses the new back-end API you've added, right?  Any reason why I couldn't easily package that up into a WAR and deploy it in the quickstart distro?

bekihm:
@EricWittmann We implemented only the EsStorage class. There are todos left for the JpaStorage class. So it relies on the elasticsearch as data storage. The quickstart works at the moment on jpa right?

manager/api/jpa/src/main/java/io/apiman/manager/api/jpa/JpaStorage.java
```diff
@@ -347,6 +325,14 @@ public void updatePlugin(PluginBean pluginBean) throws StorageException {
         super.update(pluginBean);
     }
 
+    /**
+     * @see io.apiman.manager.api.core.IStorage#updateDeveloper(DeveloperBean)
+     */
+    @Override
+    public void updateDeveloper(DeveloperBean developer) throws StorageException {
+        // TODO: Implement
```
volkflo, line 333: @EricWittmann You can find it here. As we use only elasticsearch we had not the time to implement the other storage layer as well. For the quickstart, we must implement this.

EricWittmann:
Ah right - you mentioned that a couple times already.  :)  So perhaps someone will be motivated to contribute SQL support for this feature!  And I see you've put that on the roadmap already.  Got it.

PR #918, opened 2020-10-13
Apiman Issue 784

### LINKED ISSUE #784: Manager REST API - Securing CORS by checking Origin
A recent security audit on our Apiman deployment reported vulnerabilities that I would like to share with you.

Manager REST API accept CORS requests without any controls on origin. All requests are allowed.

An example : A POST request to create an organization via manager api with a **faked Origin header** :
Apiman - manager api cors attack

We can see in [ApimanCorsFilter.java](https://github.com/apiman/apiman/blob/master/common/servlet/src/main/java/io/apiman/common/servlet/ApimanCorsFilter.java) that Origin header is automatically return in "Access-Control-Allow-Origin".

Proposal :
Check that the Origin present in the http header belongs to a list of authorized domains. This list must be configurable.

This could be done with minimal changes inspired by this Github filter project : [eBay/cors-filter](https://github.com/eBay/cors-filter). It covert all CORS uses cases, including Origin checking. And when a request is invalid, replying with an HTTP 403 error code.

Filter implementation has advantage to be compatible with Tomcat and WildFly. Otherwise, maybe with JAX-RS ContainerResponseFilter.

alograg:
Implementation of code to define the headers of the cross-origin request (CORS) in server initialization command.
Ex:
`./bin/standalone.sh -b 0.0.0.0 -bmanagement 0.0.0.0
   -c standalone-apiman.xml
   -Dapiman-manager-ui.allowed-cors-origins=localhost,alograg.me
`
Fixes #784


alograg:
How can i success security/snyk check?

SimonClavier:
> 
> 
> How can i success security/snyk check?

I think we should ask @EricWittmann for help.
On my side I can't access the details on the Security/Snyk Page, I have the following error message :
_"Unable to display this organization
The organization does not exist, or you do not have permissions to access it."_

EricWittmann:
I actually don't have access to the snyk org, I think @msavy may have set that up.  I'm inclined to disable that check since I've been ignoring it for ... ever.

EricWittmann:
That said, there are conflicts in this PR now (perhaps due to a recent merge) - so have a look at those (I haven't checked yet) while we figure out what to do about Snyk.  :)

volkflo:
Yeah, pretty sure because of the developer portal backend. I implement more or less the same feature but I did not remember that there was an open issue with that 👀

msavy:
Eric, I added you to the snyk org a while back. It should have sent you an invite. I can try getting it to send again if that helps 

Makes sense to disable it now probably as I think GitHub provides enough OOTB probably.

alograg:
@volkflo can you review my code?

volkflo:
Hi @alograg,

as I mentioned above: Our implementations are more or less the same. they just differ a little bit in how we get the system property.
You can compare it here: https://github.com/apiman/apiman/blob/5c85f0e1a19e13804f7dc2a1634ac0525b6a57ff/common/servlet/src/main/java/io/apiman/common/servlet/ApimanCorsFilter.java

But also my implementation with the system property is also just a workaround.
It would be nice to read the value out of the `apiman.properties` instead.

common/servlet/src/main/java/io/apiman/common/servlet/ApimanCorsFilter.java
```diff
@@ -57,16 +57,20 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
         HttpServletRequest httpReq = (HttpServletRequest) request;
         HttpServletResponse httpResp = (HttpServletResponse) response;
 
-        if (isPreflightRequest(httpReq) && originIsAllowed(httpReq)) {
-            httpResp.setHeader("Access-Control-Allow-Origin", httpReq.getHeader("Origin")); //$NON-NLS-1$ //$NON-NLS-2$
+        // Get system property here
+        String[] allowOriginsA = System.getProperty("apiman-manager-ui.allowed-cors-origins", "*").split(",");
```
volkflo, line 61: Please compare with existing code: https://github.com/apiman/apiman/blob/5c85f0e1a19e13804f7dc2a1634ac0525b6a57ff/common/servlet/src/main/java/io/apiman/common/servlet/ApimanCorsFilter.java#L37-L42

alograg: Nice approach.
I will do something like that.
> It would be nice to read the value out of the apiman.properties instead.
I'm not a pro in JAVA, more and less code monkey :P
Where can i find an exemple of reading properties of `apiman.properties` ?
I haven't find one.

volkflo: I think a good start would be here: https://github.com/apiman/apiman/blob/master/manager/ui/war/src/main/java/io/apiman/manager/ui/server/UIConfig.java

alograg:
Hi @volkflo 
I changed the class to add the configuration inside the properties file and it can be overwritten on the command line.
I await your comments.
Thank you all, i have learned a lot.

common/servlet/src/main/java/io/apiman/common/servlet/ApimanCorsFilter.java
alograg: I have added all, your approach and the configuration.

volkflo:
I am on holidays this week. I will probably take a look on your changes next week if no one other of the maintainers will do it this week :)Am 24.10.20, 21:58 schrieb Henry <notifications@github.com>:

   
  Hi @volkflo I changed the class to add the configuration inside the properties file and it can be overwritten on the command line. I await your comments. Thank you all, i have learned a lot.

distro/wildfly/src/main/resources/overlay/standalone/configuration/apiman.properties
```diff
@@ -46,6 +46,9 @@ apiman.hibernate.hbm2ddl.auto=validate
 apiman-manager-ui.api.authentication.type=bearerToken
 apiman-manager-ui.api.authentication.token.generator=io.apiman.manager.ui.server.kc.KeyCloakBearerTokenGenerator
 
+# Set crosn origin forn the manager; *=any, comma separated names
```
bekihm, line 49: some typos: cross, for the manager

common/servlet/src/main/java/io/apiman/common/servlet/ApimanCorsFilter.java
```diff
@@ -29,16 +31,23 @@
...
      * Constructor.
      */
     public ApimanCorsFilter() {
-        String corsOrigins = System.getProperty("allowed_cors_origins");
-        if (corsOrigins != null) {
-            allowedCorsOrigins = new HashSet<>(Arrays.asList(corsOrigins.trim().split(",")));
-        }
+        String corsOrigins = System.getProperty(MANAGER_UI_ALLOWED_CORS_ORIGINS,
+                                                config.getString(MANAGER_UI_ALLOWED_CORS_ORIGINS,
+                                                                "*"));
```
bekihm, line 49: So we allow all origins by default, if config is not set?

bekihm (changes requested):

common/servlet/src/main/java/io/apiman/common/servlet/ApimanCorsFilter.java
EricWittmann: That was probably the original behavior... :)

bekihm: So we keep it :)

alograg: > So we allow all origins by default, if config is not set?

Yes

distro/wildfly/src/main/resources/overlay/standalone/configuration/apiman.properties
alograg: it's true, i will correct it

distro/wildfly/src/main/resources/overlay/standalone/configuration/apiman.properties
```diff
@@ -46,6 +46,9 @@ apiman.hibernate.hbm2ddl.auto=validate
 apiman-manager-ui.api.authentication.type=bearerToken
 apiman-manager-ui.api.authentication.token.generator=io.apiman.manager.ui.server.kc.KeyCloakBearerTokenGenerator
 
+# Set cross origin forn the manager; *=any, comma separated names
```
bekihm, line 49: 'for' instead of 'forn'

bekihm: After this, I think it is ready for merge :-)

alograg: ups, sorry again.

bekihm (approved):

PR #922, opened 2020-10-22
Rollback ddl files and force update of gateway password

volkflo:
With the commit of the devportal, I pushed some broken ddls (probably from my testing testing phase).
The change of the date in the initial data will force an update of the gateway password.

### DIFF
--- a/distro/data/src/main/resources/bootstrap/default-gateway.json
+++ b/distro/data/src/main/resources/bootstrap/default-gateway.json
@@ -1,6 +1,6 @@
 {
   "Metadata": {
-    "exportedOn": "2020-08-11T12:00:00Z",
+    "exportedOn": "2020-10-22T12:00:00Z",
     "apimanVersion": "${version}"
   },
   "Gateways": [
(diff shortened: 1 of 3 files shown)


EricWittmann:
These changes seem OK but my testing of the 2.0.0 docker image seems to show that the proper credentials for the `apimanager` user are `apimanager123!` which is different from what's in the `distro/data/src/main/resources/bootstrap/default-gateway.json` file.  We could change the default in that JSON file, no?

I'm not sure how the date change would affect the password problem.

EricWittmann:
I've re-confirmed this in the latest docker image.  The default credentials we need to use for the Gateway are:

* `apimanager`
* `apimanager123!`

The `default-gateway.json` file uses a default password value of `apiman123!` - so I think if we just change that it should fix the docker quickstart.

volkflo:
@EricWittmann For the password please see my comment here: https://github.com/apiman/apiman-docker/issues/20#issuecomment-714273453

About the date:
In https://github.com/apiman/apiman/pull/893 I changed refactored the mechanism for initial data and how the files are imported.
The date now works as an ID.
When apiman now starts the following happens:

1. read the file with the initial data and get the date
2. Check in the database if there exists a metadata entry with this "id"/date
3. If not, import the data
4. Create a new metadata entry with apiman version and timestamps

Now if we want to change update initial/system data we just need to change the data and update the date.

For the gateway this means. If you start a new version of apiman it will not have a metadata entry for this timestamp. 
This will force an import and with this an update of the gateway password.
image

EricWittmann:
I see!  Thanks for the explanation.

EricWittmann (approved):

volkflo:
@EricWittmann Maybe we can simply this liquibase stuff somehow? 
It seems that I break something accidentally every time I have to do something with the ddl files :D

EricWittmann:
I was never in favor of the liquibase approach - I think the idea was fine but I don't think it actually made anything easier.  I'd be in favor of throwing it out in favor of manually maintaining the DDLs.

PR #929, opened 2020-11-05
Push wildfly docker image on master build

volkflo:
I would like to push the docker image of wildfly on each master build.
I am not sure if this already works because I am not very familiar with github actions.

@EricWittmann Can you please take a look at it? :)

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -46,3 +46,9 @@ jobs:
         run: mvn test install -Dinvoker.skip -Dmaven.javadoc.skip=true -B -Dapiman.gateway-test.config=vertx3-file
       - name: Run Tests (AMG 1)
         run: mvn test install -Dinvoker.skip -Dmaven.javadoc.skip=true -B -Dapiman.gateway-test.config=amg-1
+      - name: Build and Publish Docker Images
+        if: github.ref == 'refs/heads/master'
+        run: |
+          mvn clean package docker:build -P docker -DskipTests
+          echo ${{ secrets.DOCKERHUB_PASSWORD }} | docker login -u ${{ secrets.DOCKERHUB_USERNAME }} --password-stdin
+          docker push apiman/on-wildfly:latest
--- a/distro/wildfly/pom.xml
+++ b/distro/wildfly/pom.xml
@@ -14,7 +14,7 @@
   <properties>
     <docker.wildfly.version>wildfly</docker.wildfly.version>
     <docker.base-image>jboss/wildfly:20.0.1.Final</docker.base-image>
-    <docker.target-image-name>apiman/on-wildfly:master</docker.target-image-name>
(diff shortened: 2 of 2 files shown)


.github/workflows/verify.yaml
```diff
@@ -46,3 +46,9 @@ jobs:
         run: mvn test install -Dinvoker.skip -Dmaven.javadoc.skip=true -B -Dapiman.gateway-test.config=vertx3-file
       - name: Run Tests (AMG 1)
         run: mvn test install -Dinvoker.skip -Dmaven.javadoc.skip=true -B -Dapiman.gateway-test.config=amg-1
+      - name: Build and Publish Docker Images
+        if: github.ref == 'refs/heads/master'
+        run: |
+          mvn clean package docker:build -P docker -DskipTests
```
EricWittmann, line 52: Have you tried this locally?  Does it work?  I didn't think we had support for building a docker image using maven.  But looking at the code I *was* able to find a dockerfile and some configuration in the pom.xml.  So I guess this works?  If you can confirm that this produces what we want, then I can add the secrets and merge this PR.  It looks good.

msavy:
This was something I contributed. It uses/used the Docker Maven Plugin to
build a master Docker image.

>
>

msavy:
See https://github.com/apiman/apiman/blob/master/distro/wildfly/src/main/docker/Dockerfile
And https://github.com/apiman/apiman/blob/master/distro/wildfly/pom.xml#L145-L198

.github/workflows/verify.yaml
volkflo: Yes, this should work after the project build has passed. But I am not sure if the `github.ref == 'refs/heads/master'` part is correct

volkflo:
> This was something I contributed. It uses/used the Docker Maven Plugin to build a master Docker image.
> […](#)

I changed this to `latest` because this more common I think.

.github/workflows/verify.yaml
EricWittmann: Yeah that should be fine - it just limits the step to the `master` branch.  Which I think is fine for us.

EricWittmann:
Yeah the change to `latest` is correct.

EricWittmann:
I'm happy with this if you want to mark is ready for review.

volkflo:
> I'm happy with this if you want to mark is ready for review.

Ready for review :)

msavy:
IMO, you should then publish to both names (or alias in some way),
otherwise people using master will end up with stale images and it won't
behave consistently.

volkflo:
@msavy For the new image (apiman/on-wildfly (without a version)) there is no master at the moment. 
See: https://hub.docker.com/r/apiman/on-wildfly/tags

msavy:
Okay, cool, thanks for that info. It would be good to add that back at some point.

volkflo:
@EricWittmann Can we try this? :)

EricWittmann (approved):

PR #936, opened 2020-11-21
FIX 932 : take basic authentication from userInfo to request the maven repository

wbeck-fr:
Use the url syntaxe like http://user:password@maven.repository to make an authentification on private nexus.
Impacted in Servlet plateform and Vertx


EricWittmann (approved):

EricWittmann:
Nice change, thanks.

Issue #919, opened 2020-10-14
Implement JPA backend for the developer portal


volkflo:
The developer portal currently works only with elasticsearch as a storage driver for the Manager UI.
To improve the overall setup and quickstart it would be nice to have also a JPA implementation.

Where to start?
https://github.com/apiman/apiman/blob/5c85f0e1a19e13804f7dc2a1634ac0525b6a57ff/manager/api/jpa/src/main/java/io/apiman/manager/api/jpa/JpaStorage.java#L176-L182

OmarMorales71:
@volkflo Hello Can I work on this? are you asking for implement DeveloperBean as a Jpa identity and create its repository?

volkflo:
@OmarMorales71 I think nobody is working on this right now, so you can work on this :)

I think there is no that much to do:

- You have to implement all methods about the DeveloperBean in `manager/api/jpa/src/main/java/io/apiman/manager/api/jpa/JpaStorage.java`
- You have to define the database tables here `https://github.com/apiman/apiman/tree/master/distro/data/src/main/resources/ddls`
- And there are also test, for e.g. `manager/test/api/src/test/resources/test-plans/exportDeveloper-testPlan.xml` there you have only to remove the `skipStorage="jpa"`

I hope this helps you a little bit :)
Feel free to ask further questions if things are not clear enough.

OmarMorales71:
@volkflo I'm working on it!

OmarMorales71:
@volkflo 
`src/main/java/io/apiman/manager/api/beans/developers/DeveloperBean.java `
private Set<DeveloperMappingBean> clients = new LinkedHashSet<>();

Can you tell more about these relationship? is OneToMany o ManyToMany? (If there is a relationship, I suppose I have to implement Jpa to DeveloperMappingBean also?)

```java
public class DeveloperBean implements Serializable {

    private static final long serialVersionUID = 7127400624541487145L;

    @Id
    @Column(nullable = false)
    private String id;
    private Set<DeveloperMappingBean> clients = new LinkedHashSet<>();
```

OmarMorales71:
@volkflo If I have to build a relationship between DeveloperMappingBean and Developer Bean I have to change the data type of the attributes of DeveloperMappingBean right?

From this
```java
public class DeveloperMappingBean implements Serializable {

    private static final long serialVersionUID = -5334196591430185705L;

    private String clientId;
    private String organizationId;
```
To this?
```java
[...]

PR #944, opened 2020-12-19
Implemented JPA for DeveloperBean. #919

OmarMorales71:
Fixes : #919 

It's the implementation JPA for DeveloperBean. Relationship many to many between Developer and Client is added.

UNIT TEST: NO

### DIFF
--- a/manager/api/beans/src/main/java/io/apiman/manager/api/beans/clients/ClientBean.java
+++ b/manager/api/beans/src/main/java/io/apiman/manager/api/beans/clients/ClientBean.java
@@ -15,6 +15,7 @@
  */
 package io.apiman.manager.api.beans.clients;
 
+import io.apiman.manager.api.beans.developers.DeveloperBean;
 import io.apiman.manager.api.beans.orgs.OrganizationBasedCompositeId;
 import io.apiman.manager.api.beans.orgs.OrganizationBean;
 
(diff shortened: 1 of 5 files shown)


bekihm:
Hello @OmarMorales71,
As a first step you can enable the export test for the jpa storage like @volkflo said in the corresponding issue:

> And there are also test, for e.g. manager/test/api/src/test/resources/test-plans/exportDeveloper-testPlan.xml there you have only to remove the skipStorage="jpa" 

After you have enable these you can try a `mvn install` which executes also some tests. An alternative is to run `mvn test`, I think.

Beware also the second hint of @volkflo :
> You have to define the database tables here https://github.com/apiman/apiman/tree/master/distro/data/src/main/resources/ddls

As a second step:
You have an error during the vertx-mem exporting tests:
`ERROR [qtp1834112243-203]: RESTEASY002025: Unknown exception while executing GET /downloads/3a0d16d1-6b35-4cac-926f-8a72efc4f949
10073
java.lang.RuntimeException: java.lang.IllegalArgumentException: org.hibernate.hql.internal.ast.QuerySyntaxException: DeveloperBean is not mapped [SELECT d FROM DeveloperBean d ORDER BY d.id ASC]
10074
	at io.apiman.manager.api.exportimport.manager.StorageExporter.exportDevelopers(StorageExporter.java:414) ~[apiman-manager-api-export-import-api-2.0.1-SNAPSHOT.jar:?]
10075
	at io.apiman.manager.api.exportimport.manager.StorageExporter.export(StorageExporter.java:90) ~[apiman-manager-api-export-import-api-2.0.1-SNAPSHOT.jar:?] `

You can run the vertx-mem tests locally with:
`mvn test install -Dinvoker.skip -Dmaven.javadoc.skip=true -B -Dapiman.gateway-test.config=vertx3-mem`

msavy:
I've already incorporated this into the v3 codebase I'm working on.

msavy:
Thanks @OmarMorales71 for this submission. This is implemented on the 3.x branch. A few things were missing (e.g. hashcode and equals, and some tests), but you were certainly along the right lines and not far off!

I hope you will consider contributing again in the future, and we greatly appreciate you being part of the community.

PR #947, opened 2021-01-04
Allow redirect from non-https origins

tevosouza:


volkflo:
Great, I will try this probably the next days 😊

distro/data/src/main/resources/data/apiman-realm-for-keycloak.json
```diff
@@ -108,17 +108,19 @@
             "enabled" : true,
             "baseUrl" : "/apimanui",
             "redirectUris" : [
-                "https://localhost:8443/apimanui/*"
+                "https://localhost:8443/apimanui/*",
+                "http://localhost:8080/apimanui/*"
             ],
             "secret" : "password",
-            "webOrigins" : [ "*" ]
+            "webOrigins" : [ "+" ]
```
msavy, line 115: Is this a reference to this? https://lists.jboss.org/pipermail/keycloak-user/2017-October/012008.html

msavy:
Is the addition of the H2 `keycloak.mv.db` necessary? Interested to hear thoughts on this. 🤔

tevosouza:
Hi @msavy,
When I did my tests to change redirect I changed only the JSON file, but it doesn't make any effect and then I put apiman up and changed config on keycloak and realized that H2 database is entirely deployed on the build and the JSON realm configuration file is not imported on stack startup.
Maybe we can change it via CLI to process realm configuration at keycloak startup but I think that we need to split apiman and master realms because keycloak base clients are exported too and it maybe causes initialization problems (i don't tested it yet...)

msavy:
Yes, it might just be a side-effect/autogeneration and possibly can be excluded safely. Give it a try when you have time 👍

distro/data/src/main/resources/data/apiman-realm-for-keycloak.json
tevosouza: Could be! heh
I made this change for security purposes, if we use '*' any weborigin is allowed to consume resources as a client, when I set to the '+' only valid redirects (set on the client configuration) is allowed as weborigins and it prevents unauthorized access and solves some CORS problem too, but CORS problem is not the case :).

volkflo:
@msavy I made the last changes on the `keycloak.mv.db` and forced everyone to use https. This file must be included because the quickstart does not automatically import the `apiman-realm-for-keycloak.json`. (Not sure if this is has worked before?).

At the moment this is more or less a template if you want to start with a standalone deployment.

msavy:
Aha, that sounds familiar, so should be ok! 

Thank you

volkflo (approved):

PR #953, opened 2021-01-15
Fix for API definition forbidden (http 401) when access from manager

tevosouza:
I faced this problem when trying to implement a soap web service for our legacy system, the apiman always returned 401 (with auth correct) when I tried to access the definition directly from the manager-ui the error still occurs and then I changed the web.xml deleting GET from organizations/* and it works, but I don't know if this is the correct/best solution.

Please, can you tell me your thoughts about it?

### DIFF
--- a/manager/api/war/wildfly8/src/main/webapp/WEB-INF/web.xml
+++ b/manager/api/war/wildfly8/src/main/webapp/WEB-INF/web.xml
@@ -101,7 +101,6 @@
     <web-resource-collection>
       <web-resource-name>apiman</web-resource-name>
       <url-pattern>/organizations/*</url-pattern>
-      <http-method>GET</http-method>
       <http-method>POST</http-method>
       <http-method>PUT</http-method>
       <http-method>DELETE</http-method>


volkflo:
I am not totally sure if I understand what you trying.
But what you are doing when you remove the `GET` here is that you removing the authentication check and this resource will be unprotected.
Are you trying to access this url?
image
If I remember it is possible that you have to enable basic auth first in keycloak otherwise the endpoint accepts only tokens.

tevosouza:
Hi @volkflo 
The problem is: when I click on the link that you printed from apiman manager I got an error 401.
I think it works in tomcat because has a AuthFilter, but Wildly doesn't not use the filter and validate all requests and causes the error. 
I saw a comment on the code removing permission check for getApiDefinition but still throws the error.

msavy:
Thanks for bringing this issue to our attention! We will try to address this through #1079.

The current solution proposed removes security checks from all GET requests which means that people may be able to access resources that they do not have permissions for by guessing the URI.

We will definitely circle around to this one shortly though.

PR #954, opened 2021-01-22
Changed gateway and gateway-api url from https to http.

tevosouza:
I made this change to allow .net developers to add a reference to APIMAN managed API direct from visual studio. I realized that the Wildfly (i didn't test on tomcat) thrusts in any certificate  ` ATTENTION: SSLSessionStrategy will trust *any* certificate. This is extremely unsafe for production. ` but if access comes from outside of Wildfly (e.g from vs) the user got an error because the server root certificate is not thrust by the client and all requests result in an SSL error.

### DIFF
--- a/distro/data/src/main/resources/bootstrap/default-gateway.json
+++ b/distro/data/src/main/resources/bootstrap/default-gateway.json
@@ -13,7 +13,7 @@
       "modifiedBy": "admin",
       "modifiedOn": "2018-05-09T12:00:00Z",
       "type": "REST",
-      "configuration": "{\"endpoint\":\"${apiman.gateway-endpoint:https://localhost:8443/apiman-gateway-api}\",\"username\":\"${apiman.gateway-endpoint.username:apimanager}\",\"password\":\"${apiman.gateway-endpoint.password:apiman123!}\"}"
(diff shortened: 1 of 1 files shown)


volkflo:
Hi @tevosouza,

I am not sure if this should be the default.
I think in production you should move to valid certificates instead of http and/or self-signed ones.

Maybe I do not understand the problem correctly :)

tevosouza:
> Hi @tevosouza,
> 
> I am not sure if this should be the default.
> I think in production you should move to valid certificates instead of http and/or self-signed ones.
> 
> Maybe I do not understand the problem correctly :)

I agree that it not can be the default configuration for the production setup , I made this change thinking on mitigate possible problems on quickstart for new users or some average users. 
If you make a request to gateway using curl or another tool like soap-ui (depend on config) or visual studio it will fail because certificate (localhost) is not accepted as a valid certificate. This PR change this behavior to allow access from HTTP.
That's the idea ;)

As C# developer I made few tweaks on the configuration to allow the visual studio to add the service reference to an apiman proxied API and it's not too trivial for new users.

msavy:
If more discussion is required, then please reopen. Our current preference is to ship with https by default so that it's more secure, and allow people to change to plaintext only if needed. 

Admittedly this means there can be complaints about certificates. If this turns out to be a significant hindrance, we should figure out a more user-friendly way to work with certs/https.

PR #956, opened 2021-01-29
Make elasticsearch logging output clearer

volkflo:
@svzieg and I needed a while to figure out that we wrongly configured `client.host`. So we made the log it a little bit clearer.


bekihm (approved):
LGTM

PR #968, opened 2021-02-15
Importer should log file name

BastianGem:
### Changes to the Importer
The importer now logs the filename at startup.

### Changes to the test framework
To test variable content in text payloads, regular expressions were enabled.

For this jUnits AssertEqual was replaced by the Java regex compiler.
In rest tests with text payload, special characters must be escaped for the regex compiler.

### DIFF
--- a/manager/test/api/src/test/resources/test-plan-data/import/001_import.resttest
+++ b/manager/test/api/src/test/resources/test-plan-data/import/001_import.resttest
@@ -370,7 +370,7 @@ Content-Type: application/json
 Content-Type: text/plain;charset=utf-8
 
 INFO: ----------------------------
-INFO: Starting apiman data import.
+INFO: Starting apiman data import: .*apiman_import_migrated.*.json
 INFO: Importing a user: importeduser
 INFO: Importing a gateway: The Gateway
 INFO: Importing an organization: Organization 1
--- a/manager/test/api/src/test/resources/test-plan-data/import/003_import-with-developer.resttest
+++ b/manager/test/api/src/test/resources/test-plan-data/import/003_import-with-developer.resttest
@@ -31,7 +31,7 @@ Content-Type: application/json
 Content-Type: text/plain;charset=utf-8
 
 INFO: ----------------------------
-INFO: Starting apiman data import.
+INFO: Starting apiman data import: .*apiman_import_migrated.*.json
 INFO: Importing a developer: Developer1
 INFO: Importing a developer: Developer2
 INFO: Publishing APIs to the gateway.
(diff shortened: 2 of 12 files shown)


volkflo (approved):
LGTM :)

bekihm (dismissed):
LGTM

msavy (approved):

msavy:
@BastianGem I can't push to this branch, so [I'll just put a branch for you here](https://github.com/msavy/apiman/commits/ImporterShouldLogFileName) and you can cherry-pick the commit into the code yourself when you are ready (I also rebased it, so you might just want to push the branch over the top of your existing one). 

Another small note -- please reword your commit message to use the conventional commits pattern. You can see examples from the Apiman git log (see also https://www.conventionalcommits.org/en/v1.0.0/). 

Something like:

```
fix(importer): log file name when importing 

<If you want, a longer description, including description of the change you made to test harness>

Fixes #<some ticket number if there is one>
```

manager/test/api/src/test/resources/test-plan-data/import/001_import.resttest
```diff
@@ -1,403 +1,403 @@
...
+      ]
+    }
+  ]
+}
+----
+200
+Content-Type: text/plain;charset=utf-8
+
+INFO: ----------------------------
+INFO: Starting apiman data import: .*apiman_import_migrated.*.json
```
msavy, line 373: I think there are some weird Windows line-endings in the original file that is causing the whole file to appear to be changed? If you can find a way of leaving that alone and only making your down change that would be great great -- but don't spend much time on it!

manager/test/api/src/test/resources/test-plan-data/import/003_import-with-developer.resttest
```diff
@@ -31,7 +31,7 @@ Content-Type: application/json
 Content-Type: text/plain;charset=utf-8
 
 INFO: ----------------------------
-INFO: Starting apiman data import.
+INFO: Starting apiman data import: .*apiman_import_migrated.*.json
```
msavy, line 34: This regex match is definitely a useful feature 👍

test/common/src/main/java/io/apiman/test/common/util/TestPlanRunner.java
```diff
@@ -34,6 +34,8 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.regex.Matcher;
```
volkflo, line 37: We do not need this anymore because of @msavy's assert

test/common/src/main/java/io/apiman/test/common/util/TestPlanRunner.java
```diff
@@ -45,6 +47,7 @@
 import org.custommonkey.xmlunit.XMLAssert;
 import org.custommonkey.xmlunit.XMLUnit;
 import org.junit.Assert;
+import org.junit.ComparisonFailure;
```
volkflo, line 50: We do not need this anymore because of @msavy's assert

volkflo (changes requested):

volkflo (approved):

PR #973, opened 2021-02-17
refactor: deps and core code updated to support up to Java 15

### LINKED ISSUE #796: Support Java 11+
Right now I think we're stuck on Java 8.  Let's make sure we can build and run using Java 11!

msavy:
* Code updated to allow the project to build and run on versions of Java after 8, and minimally up to Java 15. Much of this pertains to the Java module system changes introduced in Java 9, but also required some major refactoring of LDAP tests as the Apache LDAP library required a major version upgrade.

* Some changes in security policy meant that tests needed updating to have larger RSA keysizes, for example. In newer releases of the JRE, including certain patch releases, RSA keys shorter than 512 bits have a been banned by an updated `java.security` policy.

* Some older platforms may still not support newer versions of Java, but this is not a fault with Apiman (e.g. old versions of Wildfly).

* Refactored LDAP tests to reduce code duplication, as it was very hard to see what the differences were between LDAP and LDAPS tests, which was causing some issues during refactoring.

* Added a method to allow the `SSLSocketFactory` to be overridden in `ILdapComponent`, which is useful for testing and perhaps certain production scenarios. Removed workaround/hack from tests.

Fixes #796


gateway/engine/core/src/main/java/io/apiman/gateway/engine/impl/DefaultLdapSearchImpl.java
```diff
@@ -66,16 +68,13 @@ private void getResults(String searchDn, String filter, LdapSearchScope scope,
...
-
-            @Override
-            public void handle(IAsyncResult<List<SearchResultEntry>> results) {
-                if (results.isSuccess()) {
-                    List<ILdapSearchEntry> searchResults = toSearchEntry(results.getResult());
-                    resultHandler.handle(AsyncResultImpl.create(searchResults));
-                } else {
-                    resultHandler.handle(AsyncResultImpl.<List<ILdapSearchEntry>>create(results.getError()));
-                }
+        System.out.println("Doing a search....");
```
msavy, line 71: Stray println statement

pom.xml
```diff
@@ -1346,15 +1355,15 @@
...
+<!--            <dependency>-->
+<!--              <groupId>io.vertx</groupId>-->
+<!--              <artifactId>vertx-platform</artifactId>-->
+<!--              <version>${version.io.vertx}</version>-->
+<!--            </dependency>-->
             <dependency>
               <groupId>io.vertx</groupId>
               <artifactId>vertx-hazelcast</artifactId>
-              <version>${version.io.vertx}</version>
+              <version>3.7.0</version>
```
msavy, line 1366: Either remove to change to property-defined version.

pom.xml
```diff
@@ -1474,6 +1483,9 @@
             </goals>
           </execution>
         </executions>
+        <configuration>
+          <source>8</source>
```
msavy, line 1487: Think IDE might have added this, either way probably not needed?

msavy:
Looks like a couple of small issues with ES. Shouldn't be hard to fix, just a dep upgrade by the looks of it. 

Stand by...

msavy:
The CI build action seems to be stuck and it won't cancel!

gateway/platforms/vertx3/vertx3/src/test/resources/io/apiman/gateway/platforms/vertx3/users.ldif
```diff
@@ -1,23 +1,28 @@
 dn: ou=people,o=apiman
+changetype: add
```
volkflo, line 2: Can you explain what this is for?

gateway/test/src/test/resources/vertx3/conf-es.json
```diff
@@ -181,5 +181,3 @@
   // When reporting a API's endpoint, prefer secure (i.e. https)
   "preferSecure": false
 }
-
-}
```
volkflo, line 185: Is this okay?

volkflo (commented):
I am not very into the whole LDAP stuff so I accepted the most of it :D 
But added some comments :)

gateway/test/src/test/resources/vertx3/conf-es.json
msavy: There was a syntax error in the file before with an unmatched end brace; it was evidently being ignored in previous versions for whatever reason. 

If you check the file, you will see the JSON syntax is now valid.

I have no idea why it worked previously.

gateway/engine/core/src/main/java/io/apiman/gateway/engine/impl/DefaultLdapSearchImpl.java
```diff
@@ -58,6 +56,10 @@ private void getResults(String searchDn, String filter, LdapSearchScope scope,
             List<SearchResultEntry> searchResults = connection.search(searchDn, searchScope, filter).getSearchEntries();
             result.handle(AsyncResultImpl.create(searchResults));
         } catch (LDAPException e) {
+            if (ldapErrorHandler == null) {
+                // TODO wire in logger
+                System.err.println("LDAP Error Handler not set. Error may be swallowed; this is probably not what you intended.");
+            }
```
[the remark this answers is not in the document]
msavy: Apparently if you click resolve with a comment, the comment is not saved. 

Anyway, second try...

Yes, logging is tricky because we try to work on a variety of platforms without interfering with the underlying logging config. We need to try to crack it. I left the TODO so we can sweep it up later.

gateway/platforms/vertx3/vertx3/src/main/generated/io/apiman/gateway/platforms/vertx3/common/config/InheritingHttpServerOptionsConverter.java
```diff
@@ -158,6 +158,16 @@ public static void fromJson(Iterable<java.util.Map.Entry<String, Object>> json,
...
+        case "maxWebSocketFrameSize":
+          if (member.getValue() instanceof Number) {
+            obj.setMaxWebSocketFrameSize(((Number)member.getValue()).intValue());
+          }
+          break;
+        case "maxWebSocketMessageSize":
+          if (member.getValue() instanceof Number) {
+            obj.setMaxWebSocketMessageSize(((Number)member.getValue()).intValue());
+          }
+          break;
```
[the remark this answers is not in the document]
msavy: Lost the comment I put here previously, trying again...

I don't have any objection to checking in this sort of generated configuration code. It's been checked in for the last few years, so it probably makes sense to be consistent.

If you think this is a problem, let's open a ticket and we can consider explicitly removing it - I don't think there would be a problem, but there could be some unintended consequences we haven't thought about.

gateway/platforms/vertx3/vertx3/src/test/resources/io/apiman/gateway/platforms/vertx3/users.ldif
msavy: The changetype attribute is now explicitly required by the underlying LDAP library used for testing, otherwise it will fail.

changetype tells the LDAP engine what sort of change you want to make. for example, you are adding, deleting, modifying, etc.

See:
https://www.digitalocean.com/community/tutorials/how-to-use-ldif-files-to-make-changes-to-an-openldap-system

msavy:
@volkflo thanks. the more eyes, the better!

volkflo:
@msavy can you also change the pipeline to use java 11+? :)
https://github.com/apiman/apiman/blob/d8217bf48349454f95ebe4c284c526de942e9874/.github/workflows/verify.yaml#L28-L31

msavy:
Yes, I think we should do Java 8, 11, and latest if possible. 

I was thinking we should make another pr, but we can do it here if you prefer. What are your thoughts?

gateway/engine/core/src/main/java/io/apiman/gateway/engine/impl/DefaultLdapSearchImpl.java
```diff
@@ -16,6 +16,10 @@
 
 package io.apiman.gateway.engine.impl;
 
+import com.unboundid.ldap.sdk.LDAPConnection;
```
msavy, line 19: fix ordering of imports (grr, intellij!(

distro/wildfly/src/main/resources/overlay/standalone/configuration/apiman.properties
```diff
@@ -20,6 +20,7 @@ apiman-gateway.jdbc.jndi-location=java:jboss/datasources/apiman-gateway
 # The following are settings for using elasticsearch for various
 # apiman components.
 # ---------------------------------------------------------------------
+#apiman.es.client.type=es
```
volkflo, line 23: If I am not wrong this will also change the quickstart back to Elasticsearch?

gateway/engine/policies/pom.xml
```diff
@@ -113,6 +113,13 @@
           </instructions>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <argLine>-Xmx5g</argLine>
```
volkflo, line 120: Do we really need such a big max value for memory?

pom.xml
```diff
@@ -1336,37 +1345,14 @@
           <artifactId>docker-maven-plugin</artifactId>
           <version>${version.docker.plugin}</version>
         </plugin>
-        <plugin>
-          <groupId>io.vertx</groupId>
```
volkflo, line 1340: Why do you remove the vert.x stuff here?

volkflo (commented):
Just added some questions? :)

distro/wildfly/src/main/resources/overlay/standalone/configuration/apiman.properties
msavy: No, this is just a comment so that if the user uncomments that ES configuration section it will work correctly -- it simply now uses a variable instead of repeating "jest" which is deprecated or doesn't work. 

It's a bit clearer to see in the full diff -- https://github.com/apiman/apiman/pull/973/files/94d760b27e00001ef02a9313ae25f7e3bee64a84#diff-e46e7e0fd0cbfc0b46a9b62be423440ddad0945d451078028cde91ceb75585fe

gateway/engine/policies/pom.xml
msavy: Yes, it was OOMing on CI because the Apache LDAP server used during testing spikes the heap size briefly. I reached this number after trying a few different values. This is just a ceiling number (mx = heap memory maximum), and the JVM will only use that much if it needs to.

pom.xml
msavy: Basically, it's a relic 🦕.

That plugin is no longer used since older versions of Vert.x and this was just hanging around doing nothing! 

The `maven-shade-plugin` Maven plugin now to create the Uberjar 👍. In fact, the latest version of that plugin was released in 2015 and nobody uses it any more - https://mvnrepository.com/artifact/io.vertx/vertx-maven-plugin .

msavy:
Thanks for your additional feedback and questions 💪

pom.xml
volkflo: Okay, then it is great to get rid of it :)

distro/wildfly/src/main/resources/overlay/standalone/configuration/apiman.properties
volkflo: Ahh yes, I did not recognize that these lines are all comments 🚀

msavy:
Ready to merge, someone please give me the green tick when ready :-).

gateway/engine/policies/src/test/java/io/apiman/gateway/engine/policies/BasicAuthLDAPSTest.java
```diff
@@ -14,63 +14,60 @@
...
-import java.util.HashSet;
-import java.util.Set;
 
 /**
  * Fires up an ldaps server to do a better job testing the basic auth
  * policy when configured for ldaps.
  *
  * @author jhauray
  */
+@Ignore
```
volkflo, line 65: Do we have an issue bringing this test back at some point?

volkflo (approved):

gateway/engine/policies/src/test/java/io/apiman/gateway/engine/policies/BasicAuthLDAPSTest.java
msavy: Done https://github.com/apiman/apiman/issues/994

PR #980, opened 2021-02-18
Fix es index mapping

BastianGem:
### SearchingTest
Added some Tests for API-Search

### Changes to the IndexMapping
Replaced all text-type mappings with keyword-type.
Added a normalizer to lowercase all name fields (case insensitive search)


BastianGem:
In my opinion, we should completely define the index mapping for each index in any case, this way we prevent unpredictable behavior by ElasticSearch and the default mapping.

For the actual index mapping we still have to decide how strings should be mapped.
For this we have 4 possibilities:
* Map all strings as text.
* Map all strings as keyword.
For this, a lowercase normalizer must be stored for all fields to enable a case-insensitive search.
* Map all strings as text with multi field keyword.
* Map all strings as keyword with multi field text

Excluded from the mapping possibilities are IDs, these should always be mapped as keyword (as-is).


To stay as flexible as possible and hopefully future-proof we should decide for a mapping with multi fields.
Since the text-type does not allow sorting and aggregation by default and the full text search is not part of our standard use-case I would tend to option 4.

common/es/src/main/java/io/apiman/common/es/util/EsIndexMapping.java
```diff
@@ -26,249 +28,441 @@
...
         Map<String, Object> textTypeProperty = new HashMap<>();
-        textTypeProperty.put("type", "text");
+        Map<String, Object> keywordTypeProperty = new HashMap<>();
+        Map<String, Object> multiFieldProperty = new HashMap<>();
+
+        textTypeProperty.put("type", EsConstants.ES_MAPPING_TYPE_TEXT);
         textTypeProperty.put("fielddata", true);
-        addFieldMapping(fieldName, textTypeProperty, indexFieldProperties);
+        keywordTypeProperty.put("type", EsConstants.ES_MAPPING_TYPE_KEYWORD);
+        keywordTypeProperty.put("ignore_above", 256);
```
volkflo, line 407: @BastianGem Are these lines the default values (e.g. the 256)?

BastianGem: @volkflo Yes these are deafult values.
The ignore_above property avoids indexing of huge terms, with the current setting of 256 only strings with a maximum length of 256 will be indexed

manager/test/api/src/test/resources/test-plans/searching-testPlan.xml
```diff
@@ -1,46 +1,69 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
msavy, line 1: Nit: it's hard to see what's changed here because the indentation has changed. Would it be possible to change that back so it's a little clearer in the diff?

common/es/src/main/java/io/apiman/common/es/util/EsConstants.java
```diff
@@ -78,72 +80,324 @@
 
 
     public static final String[] MANAGER_INDEX_POSTFIXES = {
-            INDEX_MANAGER_POSTFIX_GATEWAY,
```
msavy, line 81: Nit: indentation messing up the diff a bit, please match old indentation if possible!

manager/test/api/src/test/resources/test-plan-data/searching/queries/apis/010_query-by-name-lower.resttest
```diff
@@ -0,0 +1,85 @@
+POST /search/apis admin/admin
```
msavy, line 1: Good to have some extra coverage 👍

msavy (changes requested):
Thank you for this!

My immediate reaction to this is that it should work, but I reckon we should reconsider the implementation pattern used here.

Let me explain:

Most of the code that defines the ES mappings is now centralised in a single location. That means the single location needs to "know" about the existence and implementation details of those components, which tends to make it more effort to maintain.

Without much additional effort we could use a registration pattern, where each ES component registers its own mappings. This means each component remains responsible for its own schema, and is cleanly encapsulated. 

So, for example, ESMetrics is responsible for maintaining its own mappings.

If it's necessary to have single location responsible for doing this, we can create a static registry class to which we publish all of the mappings, and it can deal with doing the work of piecing stuff together.

Let's have a quick discussion, because I think this would be a small amount of effort and improve the code a lot. It would reuse the existing effort mostly, and is more a design thing.

The other alternative is that we can commit as-is, and refactor later.

manager/test/api/src/test/resources/test-plans/searching-testPlan.xml
msavy: BTW, you may need to look at the full diff to see what I mean. It doesn't show much on the PR page.

msavy:
I would love to hear if you have any thoughts on whether any of these changes will break backwards compatibility for existing users who have data in the gateway? 

For example, if they have a rate limit stored in a previous version, will it be able to resume from the existing figures after they upgrade?

msavy:
We had a productive conversation and sketched out some ideas for how we could solve this in the way I suggested in my review comments. @BastianGem and I will work on that and report back here! The solution is rather simple, so should not take much effort.

In short: 

* We will create a simple builder that allows us to quickly and efficiently produce the required index definition (in Java), and then marshall it into JSON.
* Each ES component will be responsible for maintaining its own index definition.
* Each ES component will automatically register itself with a (static) component registry `EsComponentRegistry`.
* During the ES client initialisation process, we will iterate through the ES component registry and initialise all of the index configs. 

@BastianGem also discovered a couple of issues with his code relating to the "data" segments.

msavy:
Superseded by #1023

PR #965, opened 2021-02-05
Little ES improvement

volkflo:
I probably found some unused code.
I think we can remove these lines because `truststore` is not used and a few lines below we initialize the SSL Context:
https://github.com/apiman/apiman/blob/56275fcfbfa014006c24d0acf211ef7c8ed39316/common/es/src/main/java/io/apiman/common/es/util/DefaultEsClientFactory.java#L243-L246

PR #990, opened 2021-02-23
ci: build for Java 8, 11, and 15

### LINKED ISSUE #978: Add Java 11 and Java 15 builds to CI scripts

msavy:
Fixes #978

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -21,14 +21,17 @@ jobs:
   build-verify:
     name: Verify Build
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [ 8, 11, 15 ]
     if: github.repository_owner == 'Apiman'
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
-      - name: Set up JDK 1.8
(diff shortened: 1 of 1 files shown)


volkflo (dismissed):
LGTM :) I like how simple GitHub solved this to use multiple versions of Java 🚀 
The problem while pulling an elastic image seems to be fixed now?

msavy:
@volkflo Thanks for the feedback!

I think the ES image must be permissions related, because it doesn't complain when people with commit permissions do it. So the fix may still be needed for non-committers 🤔. May need a little investigation. I suspect a few little tweaks are all that is needed.

volkflo:
@msavy I am not sure if it is a permission problem. I am also allowed to commit directly in this repository and the actions for my PR failed :D https://github.com/apiman/apiman/pull/965

volkflo:
Seems that the CI will not start your build?
image

msavy:
I noticed that yesterday. I think it's a glitch. I'll kick it again once
the other PR is in 👍. It isn't even showing in the actions menu, so
something has gone wonky in GitHub.

>
I tested the syntax locally and it works ok

PR #1001, opened 2021-02-24
ci: deploy successful master builds as SNAPSHOT to ossrh

### LINKED ISSUE #996: Automatically deploy snapshot of latest successful master build to OSSRH
Under Travis CI when there was a successful build, we'd deploy a new snapshot to OSSRH. We're not doing that any more with GitHub actions. 

We should add that back because the Apiman Plugins can't build successfully on CI any more when they require the latest Apiman snapshot updates.

I should still have the credentials I need saved somewhere (hopefully) to inject into the secrets.

https://docs.github.com/en/actions/guides/publishing-java-packages-with-maven

msavy:
When a build to master succeeds, deploy a snapshot to OSSRH. This is what we did previously when using Travis CI. We need this to ensure that apiman-plugins continues to work correctly (it builds off these snapshots).

Fixes #996

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -31,6 +31,9 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: ${{ matrix.java }}
+          server-id: ${{ secrets.OSSRH_ID }}
+          server-username: ${{ secrets.OSSRH_USERNAME }}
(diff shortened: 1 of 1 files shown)


volkflo:
@msavy Even if this is already merged I am going through your changes and try to follow what you did yesterday 🚀

PR #1007, opened 2021-02-25
build(deps): bump ant to 1.10.9 & manage version in root pom

### LINKED ISSUE #1005: Update Ant plugin and define version into root pom
https://github.com/apiman/apiman/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc+ant

msavy:
Fixes #1005

Also fixes some security-related issues with Ant, but it does not appear they are relevant for our use-cases, as we only use Ant during build time (no runtime stuff). 

In theory, there are also some permissions changes it could make in $TMP, but they would require some real contortion to exploit. 

Anyway, better to fix it and not wonder :-).

### DIFF
--- a/manager/ui/war/pom.xml
+++ b/manager/ui/war/pom.xml
@@ -186,13 +186,6 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-antrun-plugin</artifactId>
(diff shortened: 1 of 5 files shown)


volkflo (approved):
Worked on my local machine :)

PR #1011, opened 2021-02-26
test: Autostart Elastic Testcontainer for Vert.x ES integration tests

### LINKED ISSUE #1010: Use ES Testcontainer for Vert.x gateway integration tests
Rather than relying on ES being present on CI or the system, use an ES Testcontainer for Vert.x gateway integration tests. The lifecycle is managed in the same way by the servlet implementation.

msavy:
Rather than relying on a background ES instance, a factory manages the lifecycle of an Elasticsearch Testcontainer with randomised ports for the Vert.x Gateway integration testsuite.

This means that the environment does not need an ES instance running when executing these integration tests from Maven, IDE, CI, etc.

Run as `mvn test -Dapiman.gateway-test.config=vertx3-es`

Fixes #1010

### DIFF
--- a/common/es/src/main/java/io/apiman/common/es/util/DefaultEsClientFactory.java
+++ b/common/es/src/main/java/io/apiman/common/es/util/DefaultEsClientFactory.java
@@ -92,12 +92,12 @@ public class DefaultEsClientFactory extends AbstractClientFactory implements IEs
      */
     private RestHighLevelClient createEsClient(Map<String, String> config, String indexNamePrefix, List<String> defaultIndices) {
         String host = config.get("client.host"); //$NON-NLS-1$
-        Integer port = NumberUtils.toInt(config.get("client.port"), 9200); //$NON-NLS-1$
+        int port = NumberUtils.toInt(config.get("client.port"), 9200); //$NON-NLS-1$
         String protocol = config.get("client.protocol"); //$NON-NLS-1$
         String initialize = config.get("client.initialize"); //$NON-NLS-1$
         String username = config.get("client.username"); //$NON-NLS-1$
         String password = config.get("client.password"); //$NON-NLS-1$
-        Integer timeout = NumberUtils.toInt(config.get("client.timeout"), 10000); //$NON-NLS-1$
+        int timeout = NumberUtils.toInt(config.get("client.timeout"), 10000); //$NON-NLS-1$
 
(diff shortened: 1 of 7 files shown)


msavy:
p.s. until this is merged in the Vert.x tests will not pass.

common/es/src/main/java/io/apiman/common/es/util/DefaultEsClientFactory.java
```diff
@@ -92,12 +92,12 @@ public RestHighLevelClient createClient(Map<String, String> config, String defau
      */
     private RestHighLevelClient createEsClient(Map<String, String> config, String indexNamePrefix, List<String> defaultIndices) {
         String host = config.get("client.host"); //$NON-NLS-1$
-        Integer port = NumberUtils.toInt(config.get("client.port"), 9200); //$NON-NLS-1$
+        int port = NumberUtils.toInt(config.get("client.port")); //$NON-NLS-1$
```
volkflo, line 95: Is there a reason why you removed the default port?

volkflo (approved):

common/es/src/main/java/io/apiman/common/es/util/DefaultEsClientFactory.java
msavy: Ah yes, I need to change this back. 

I put it there because the behaviour of `NumberUtils.toInt` was causing confusing behaviour. When a variable substitution from system properties was meant to be happening (but was failing and remaining as text because it had not yet been set), `NumberUtils.toInt` was replacing it with 9200 rather than throwing an error.

i.e. `test.FOO_BAR` was meant to be replaced with the randomised port, but it was staying as `test.FOO_BAR`; that caused a silent error in `NumberUtils.toInt` and it was changing to 9200 rather than throwing an error.

I reckon that if the user provides no input, then a default makes sense. However, if they *do* provide an input but it is erroneous (e.g. text when it should be a number), we should throw an error rather than using a default.

I'll change it back (the alteration was just for testing), but we should open a ticket for that and change it slightly. We can make some convenience functions for this to better behave as we want.

msavy: #1012 👍

msavy:
Thanks for your review @volkflo - I've changed it back as suggested 🙏. Thank you for spotting that.

volkflo (dismissed):

PR #1031, opened 2021-03-09
fix: Delete Query for plans and clients

volkflo:
Fix queries to delete plans and clients.
For the clients, it is the same query as for APIs.
For the plans, we just fixed two copy/paste errors.

We have/should add tests for this once the other PR for the index creation is merged (because the test plans a refactored in there)


msavy:
Yes, I'm happy to have the additional tests as a separate PR. Sounds fine to me. Thanks for this.

volkflo:
Closed because of #1041

PR #1036, opened 2021-03-09
Fix es index mapping 2

BastianGem:


msavy:
Thanks, you work from this is on master now.

PR #1035, opened 2021-03-09
Add additional tests #1031

volkflo:
Just a test (WIP).
CI should fail with this.

PR #1041, opened 2021-03-10
fix(manager): add test for delete (clients, apis, plans)

volkflo:
This is a new PR that combines https://github.com/apiman/apiman/pull/1031 and https://github.com/apiman/apiman/pull/1035


manager/test/api/src/test/resources/test-plan-data/clients/delete/002_delete-client-1.resttest
```diff
@@ -1,3 +1,3 @@
-DELETE /organizations/Organization1/clients/Client1 admin/admin
-----
-204
+DELETE /organizations/Organization1/clients/Client2 admin/admin
```
BastianGem, line 1: Why did you change this plan to delete 'Client 2' instead of 'Client 1'?

manager/test/api/src/test/resources/test-plans/clients-testPlan.xml
```diff
@@ -67,5 +67,6 @@
   <testGroup name="Deleting Clients (With subordinate elements)">
     <test name="Delete Client 1">test-plan-data/clients/delete/002_delete-client-1.resttest</test>
```
BastianGem, line 68: We should decide on a consistent start of test numbering

manager/test/api/src/test/resources/test-plans/apis-testPlan.xml
```diff
@@ -143,6 +143,7 @@
     <test name="Retire API 1">test-plan-data/apis/publish/005_retire-api-1.resttest</test>
     <test name="Delete API 1">test-plan-data/apis/publish/006_delete-api-1.resttest</test>
     <test name="Delete API 1 (404)">test-plan-data/apis/delete_unpublished/004_get-unpublished-api1-404.resttest</test>
+    <test name="Check delete do not affect other APIs">test-plan-data/apis/versions/011_get-api-5-version-7.0.resttest</test>
```
BastianGem, line 146: Maybe we could think about reorganizing some testplans.
In my opinion its a little bit confusing if more than one folder and test numbering appears in one testGroup
But i think this belongs to another issue/PR

manager/test/api/src/test/resources/test-plan-data/clients/delete/002_delete-client-1.resttest
volkflo: Because Client 2 does not have any Client-Versions. So I can check if Client 1 keep all Versions after the deletion of Client 2

manager/test/api/src/test/resources/test-plans/clients-testPlan.xml
volkflo: Oh, interesting! I didn't notice that they start here with 002. I will rename them all.

manager/test/api/src/test/resources/test-plans/apis-testPlan.xml
volkflo: Agree. I think the whole tests need refactoring at some point

msavy (approved):
It looks good to me. Thanks for the improvements.

PR #1045, opened 2021-03-12
fix(gateway-vertx): debug with java 9+

volkflo:
There is a change with java 9+ about the debugger.
This change allows debugging docker containers.


msavy:
Does this still work OK with Java 8?

volkflo:
> Does this still work OK with Java 8?

No 🤣 I added `0.0.0.0`, this is working for all java versions now. (I tested it.)

msavy:
Great, I'll check it out soon!

msavy (approved):

msavy:
Thanks for this 👍

PR #1047, opened 2021-03-16
fix: Set PolicyFailure error code if '0' before write

### LINKED ISSUE #1046: Response code in failure JSON body does not always match HTTP response code
As reported by @volkflo, the body of the policy failure error sometimes has '0' as the response code, whereas the HTTP Response Code header is correctly set as, for example, 404. 

These should match.

msavy:
Fixes #1046


volkflo (approved):
LGTM, and thanks for putting a comment in there

PR #1057, opened 2021-03-19
refactor: improved logging across Apiman (core)

msavy:
Proof of concept to fiddle with (Vert.x. gateway only for now).

You can set the log implementation with system property `apiman.logger-delegate=<name>`. Logger implementations just need to provide an `@ApimanLoggerFactory(name=<name>)` annotation (needs to be under `io.apiman.common.logging` namespace, though).

Code does not currently do anything intelligent to allow auto-selection of certain logging implementations and/or graceful error handling. That can come later if we decide this seems a good option.

I've altered the default log4j2 config on the Vert.x distro to use the JSONLayout, but in reality we probably want to allow users to point to a config instead or similar.


msavy:
I reckon bumping the manager to use log4j2 by default will be helpful also (it uses it in some distros anyway, seemingly). On the gateway side, I just need to make sure we have nice behaviour on all platforms, and perhaps add some code that can do some basic detection to see which logger to prefer if available?

msavy:
Once this is done, I will make a separate commit to create a logger cache (either simple LRUMap or caffeine etc) to prevent loads of new logger creations in IPolicyContext. Assuming it's needed and not already done in the logger impls themselves).

msavy:
Once this is done I'll make separate commits so we can produce a good OOTB experience for each platform we support with relation to JSON and standard logging :-).

This is going to be something slightly different for each platform and we want to make sure we get it right and don't just bundle it all into one PR.

msavy (commented):
Self-review OK

volkflo (commented):
LGTM, I hope that I understand most of it 🤣

volkflo:
BTW, is there a reason why the pipeline did not run?

msavy:
> BTW, is there a reason why the pipeline did not run?

No idea!

PR #951, opened 2021-01-14
Some fixes and minor changes

tevosouza:
I made some minor changes and fixes on the API definition screen
- Added a new modal dialog to show validation errors to the user
- Added JSON and WSDL validation before save to fix a bug on updateDefinitionWithManagedEndpoint when API is published but the definition type is not correct. (e.g. the API definition is WSDL format but the definition type is set to SwaggerJSON)
- Changed the behavior of save button it is disabled when No Api Definition is selected.
- Fixed a bug that does not allow to change API definition type to old value after save.

Screen:
image

Known Issues
I can't found any way to validate YAML (a performative way without including some new libraries).  =(
Include support to translations on my new error dialog.

Feel free to suggest changes

PS: 
I'm sorry for the number of commits.
I made a mistake when doing rebase and lost some commits, so I had to redo it.

PR #1075, opened 2021-03-22
feat(ui): validation for api definitions & user-friendly error modal

volkflo:
This is a refactoring of https://github.com/apiman/apiman/pull/951 originally contributed by @tevosouza.

I improved some small things and reverted a bit of the indentation and code styles.

@tevosouza Please feel free to comment and review my changes.

Thanks for the contribution @tevosouza and sorry for the long delay about this.
We will make sure that you will get the credits for this change even if I opened a different PR :)

@msavy and @EricWittmann please have also a look at this (if you have time).

### DIFF
--- a/manager/ui/war/plugins/api-manager/ts/api/api-def.ts
+++ b/manager/ui/war/plugins/api-manager/ts/api/api-def.ts
@@ -2,9 +2,9 @@
 /// <reference path="../rpc.ts"/>
 module Apiman {
 
- export var ApiDefController = _module.controller('Apiman.ApiDefController',
-        ['$q', '$rootScope', '$scope', '$location', 'PageLifecycle', 'ApiEntityLoader', 'OrgSvcs', 'Logger', '$routeParams', 'ApiDefinitionSvcs', 'Configuration', 'EntityStatusSvc', 'CurrentUser',
-        ($q, $rootScope, $scope, $location, PageLifecycle, ApiEntityLoader, OrgSvcs, Logger, $routeParams, ApiDefinitionSvcs, Configuration, EntityStatusSvc, CurrentUser) => {
+export var ApiDefController = _module.controller('Apiman.ApiDefController',
+    ['$q', '$rootScope', '$scope', '$location', 'Modals', 'PageLifecycle', 'ApiEntityLoader', 'OrgSvcs', 'Logger', '$routeParams', 'ApiDefinitionSvcs', 'Configuration', 'EntityStatusSvc', 'CurrentUser',
+        ($q, $rootScope, $scope, $location, Modals, PageLifecycle, ApiEntityLoader, OrgSvcs, Logger, $routeParams, ApiDefinitionSvcs, Configuration, EntityStatusSvc, CurrentUser) => {
             var params = $routeParams;
 
             $scope.organizationId = params.org;
@@ -20,8 +20,8 @@ module Apiman {
                 { "label" : "WSDL",                  "value" : "WSDL" }
             ];
 
-            var selectType = function(newType) {
-                angular.forEach($scope.typeOptions, function(option) {
+            var selectType = function (newType) {
+                angular.forEach($scope.typeOptions, function (option) {
                     if (option.value == newType) {
                         $scope.selectedDefinitionType = option;
                     }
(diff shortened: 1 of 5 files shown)


tevosouza:
> This is a refactoring of #951 originally contributed by @tevosouza.
> 
> I improved some small things and reverted a bit of the indentation and code styles.
> 
> @tevosouza Please feel free to comment and review my changes.
> 
> Thanks for the contribution @tevosouza and sorry for the long delay about this.
> We will make sure that you will get the credits for this change even if I opened a different PR :)
> 
> @msavy and @EricWittmann please have also a look at this (if you have time).

Hi @volkflo ,
Don't worry we all working to make apiman more and more powerful API management software.
Thank you for refactoring ;)

manager/ui/war/plugins/api-manager/ts/api/api-def.ts
```diff
@@ -32,39 +32,68 @@ module Apiman {
...
+                        default:
+                            return true;
+                }
+            }
+
+            $scope.saveApi = function () {
                 console.log('$scope.selectedDefinitionType.value: ' + $scope.selectedDefinitionType.value);
                 $scope.saveButton.state = 'in-progress';
 
+                if (isValidDefinition($scope.updatedApiDefinition, $scope.selectedDefinitionType.value) == false) {
```
BastianGem, line 60: I would prefer if (!isValidDefinition($scope.updatedApiDefinition, $scope.selectedDefinitionType.value)) instead of comparing the boolean value with false

volkflo:
Further cleanup and rebased on @BastianGem changes from PR-1076

msavy:
Could you put a few screenshots into this pull request? I'll also check it out tomorrow, but it would be great for all participants to see it in action :-)!

volkflo:
@msavy Sure :)
**Bad Certificate to load a Swagger definition**
image
**Bad WSDL**
image

manager/ui/war/plugins/api-manager/ts/api/api-def.ts
```diff
@@ -36,38 +36,66 @@ module Apiman {
                 EntityStatusSvc.getEntity().modifiedBy = CurrentUser.getCurrentUser();
             };
 
-            $scope.saveApi = function() {
+            let isValidDefinition = function (data, definitionType) {
```
msavy, line 39: Think we can use `const` or `$scope.` here - whichever you prefer.

And I think we should prefer `() => {}` syntax nowadays.

msavy (dismissed):
Thanks for this change, everyone!

A couple of small things, for any new functions I think we should use the more modern and safer `() =>` syntax over `function() {}` if possible. Its semantics are more 'normal' in terms of variable scoping and so forth.

The screenshots look good. 

One note of discussion, do we think a modal is better or an inline validation feedback of some sort?

EricWittmann (approved):
lgtm

manager/ui/war/plugins/api-manager/ts/api/api-def.ts
volkflo: I change this and will give this another test (I will be careful about UI changes :D).
In general the `() => {}` is not a replacement for all functions. There are some circumstances where you should avoid them: https://www.freecodecamp.org/news/when-and-why-you-should-use-es6-arrow-functions-and-when-you-shouldnt-3d851d7f0b26/

msavy:
Feel free to leave it if you don't think it's necessary.

manager/ui/war/plugins/api-manager/ts/api/api-def.ts
msavy: Feel free to leave it if you don't think it's necessary.

volkflo:
@msavy In this case it is okay :) I changed it and tested everything again :)
 Can you give the green tick and merge? 🚀

msavy (approved):

volkflo:
> One note of discussion, do we think a modal is better or an inline validation feedback of some sort?

I think for this use case the modal is okay.
If we want also inline validation feedback I guess we should replace the textarea (screenshot) with an full editor like https://github.com/codex-team/editor.js
image

PR #1076, opened 2021-03-22
fix: refresh re-register / re-publish button after changes

BastianGem:
After removing a policy from a client/api or reloading the Re-Register/Re-Publish Button should be enabled without the need to reload the page

Therefore modifiedOn and modifiedBy need to be updated.


volkflo (approved):
This is well tested and only it only restores the old functionality.

msavy (approved):
Looks good, thanks for this!

Please squash merge. 

Is there an issue for this btw? If so, please add the fixes reference to the  commit in same style as other commits 👍

volkflo:
@msavy There was no issue for this (this was an internal Scheer PAS issue)

PR #1096, opened 2021-04-08
ci: Remove Maven build flag for parallel builds

volkflo:
IMO this makes to log very hard to read if you are searching for errors because of an updated dependency.
Let me know what you think about this? :)

### DIFF
--- a/.github/workflows/verify.yaml
+++ b/.github/workflows/verify.yaml
@@ -62,7 +62,7 @@ jobs:
             ${{ runner.os }}-maven-
       
       - name: Build Project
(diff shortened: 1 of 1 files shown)


msavy:
OK, can do. It'll make the builds a fair bit slower though.

msavy (approved):

volkflo:
Mhh, hopefully not a lot slower. Or is there maybe a flag for sorting the output in the correct order?

msavy:
Merge it in and let's see what happens

PR #1099, opened 2021-04-10
test: generate test TLS setups via scripts using 2048 bit keys + SHA256

msavy:
Increasingly 1024 bit RSA keys and SHA1 are insufficient for many
setups. The newest versions of the JVM may refuse to use 1024 bit RSA
and/or SHA1 via their Java security policy files.

This update provides scripts to allow the full test setup to be
re-generated again in the future without hassle (e.g. 4096 bit keys),
as well as bumping it up to more modern key sizes.

Refer to the `generate_setup.sh` files in each test resources folder.

In a couple of instances, this commit refactors tests to remove hard
dependencies between different resource folders, where keys from one
folder were imported into another. This no longer works as these are
now fully independent and generated from scratch (and it was a bad
idea to begin with to have different test setups depend on each other!).


volkflo:
I think I still not get this completely.
Are these scripts called every run or is this a manual process when there are changes with the key size of the JVM?

msavy:
Manual process with key sizes and any other specification changes that may be needed. There's no easy way to predict in advance what the next security change will be, however it's usually only once every 3-5 years so far -- so I think a manual process is acceptable.

PR #1115, opened 2021-04-14
build: add Maven wrapper

### LINKED ISSUE #988: Add Maven Wrapper to improve build reproducibility
We can ensure that everyone is using the exact same version of Maven to prevent any unwanted surprises.

As an interesting note, future versions of Maven will be bundled with the Maven Wrapper script (aka mvnw), but we should commit it for now as it's not widely available.

In short, it requires just a couple of files and we can further improve our build reproducibility.

msavy:
Improve reproducibility by adding Maven wrapper to our repository. This enables people with different versions of Maven (or no Maven) to successfully build the project, removing quirks related to specific version of Maven (how meta!).

Closes #988


msavy (commented):
Self-review OK and builds fine for me. 

Please try out a build locally before OKing.

`./mvnw clean install`

PR #1118, opened 2021-04-14
refactor(es): revert waitUntil index refresh strategy. Use immediate.

msavy:
Let's try this out... We can likely pick and choose which is appropriate.

### DIFF
--- a/gateway/engine/es/src/main/java/io/apiman/gateway/engine/es/EsRegistry.java
+++ b/gateway/engine/es/src/main/java/io/apiman/gateway/engine/es/EsRegistry.java
@@ -95,7 +95,7 @@ public class EsRegistry extends AbstractEsComponent implements IRegistry {
             IndexRequest indexRequest = new IndexRequest(getIndexPrefix() + EsConstants.INDEX_APIS)
                     .id(id)
                     .source(JSON_MAPPER.writeValueAsBytes(api), XContentType.JSON)
-                    .setRefreshPolicy(WriteRequest.RefreshPolicy.WAIT_UNTIL);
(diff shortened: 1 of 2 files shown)


volkflo:
Immediate will not work for all UI pages because sometimes then the information is missing that you just created. 
I think we have to put a little more logic behind it :)

msavy:
The Elastic javadoc indicates that this should force an immediate refresh, whereas wait_for waits for Elastic's  refresh interval and the notifies you.

From the docs for "refresh=true" (aka immediate):

> Refresh the relevant primary and replica shards (not the whole index) immediately after the operation occurs, so that the updated document appears in search results immediately. This should ONLY be done after careful thought and verification that it does not lead to poor performance, both from an indexing and a search standpoint.
> 
> Force a refresh as part of this request. This refresh policy does not scale for high indexing or search throughput but is useful to present a consistent view to for indices with very low traffic. And it is wonderful for tests!
> 

This should work as far as I can tell? If it doesn't, perhaps we need to add a couple of less common healthcheck waits or similar.

volkflo:
I will try to test this :)

msavy:
> I will try to test this :)

Thanks!

volkflo:
@BastianGem and I tested this, looks good. Faster than before. 
I think we can merge! 🚀

msavy:
Fantastic, thanks!

Will double check I have everything covered and then merge in.

BastianGem (dismissed):
LGTM :)
Performance is definitely increased, which is very noticeable with the Import-Feature

msavy:
I'll fix the comments and then ask you guys for a green tick. Thanks

volkflo (approved):
Fantastic, we have a faster UI experience again! :)

PR #1121, opened 2021-04-19
chore(es): allow optional use of Amazon open distro images

volkflo:
This should allow to switch between Amazon OpenDistro and the official elastic images for the test runs

### DIFF
--- a/common/es/src/main/java/io/apiman/common/es/util/EsConstants.java
+++ b/common/es/src/main/java/io/apiman/common/es/util/EsConstants.java
@@ -16,6 +16,7 @@
 package io.apiman.common.es.util;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.Properties;
 
 /**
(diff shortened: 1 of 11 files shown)


test/common/pom.xml
```diff
@@ -97,5 +101,23 @@
...
+      <artifactId>testcontainers</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>elasticsearch</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.testcontainers</groupId>
+      <artifactId>elasticsearch</artifactId>
+      <scope>compile</scope>
```
msavy, line 115: I'm still not sure about including these deps twice in this way. It's very odd 😅. I will try to figure out whether it's possible to leave out one of the two.

msavy (approved):

msavy:
Thanks for this 🚀.

msavy:
I made a PR against your branch which should fix the NPE issue.

test/common/pom.xml
volkflo: Yes, this is very strange, but I was not able to build it without this.

common/es/src/main/java/io/apiman/common/es/util/EsConstants.java
```diff
@@ -244,30 +245,24 @@
...
-            java.util.Properties p = new Properties();
-            String version;
+    public static Properties getEsVersion() {
+        if (esVersions == null) {
+            InputStream is = EsConstants.class.getResourceAsStream("apiman-es.properties");
+            Properties p = new Properties();
             try {
                 p.load(is);
             } catch (IOException e) {
+                e.printStackTrace();
```
msavy, line 262: If this is an unrecoverable error, then we should wrap the `IOException` in `UncheckedUIException` and rethrow it.

If it is ignored, then I suggest you either log it at trace level or do nothing.

msavy (changes requested):

common/es/src/main/java/io/apiman/common/es/util/EsConstants.java
volkflo: I am not totally sure if I understand you correctly. So this method is only used for tests, I think a `UncheckedUIException` is even more confusing?

volkflo: Changed it :)

msavy (approved):
Thanks, this is good to go 👍🚀

PR #1147, opened 2021-04-27
refactor: use parseInt instead of ctor/new Integer

msavy:


volkflo (approved):
Changes look okay for me. 
(For others: Javadoc is removed because the links are outdated)

PR #1156, opened 2021-04-28
build(deps): remove unused deps

msavy:
We've accumulated some dependencies over time that aren't actually used in the maven module in which they are required/included.

I will try to eliminate some of these, which should reduce the size of the relevant JARs and reduce maintenance overhead.

We need to be careful as because of the dynamic inclusion of certain things it's not always possible for a simple Maven analysis to accurately determine whether a given dependency is truly required or not. So, some careful CI is necessary also!


volkflo (approved):
LGTM, CI is also fine with it.
Just one question:
Why do you prefer `io.apiman` over `${project.groupId}`?

msavy:
> LGTM, CI is also fine with it.
> Just one question:
> Why do you prefer `io.apiman` over `${project.groupId}`?

Just to be consistent and clearer. There's no expectation that we will change our group id, and most of the project states it explicitly.

PR #1193, opened 2021-05-06
feat(distro-tomcat): use log4j2 by default to enable easy unified JSON logging

msavy:
Note, I've left the JSON logging option enabled so it's easier for people to try out, but we'll ship with human-readable. I'll do a commit before we merge changing it back to 'default'.


msavy:
Thanks also to @outofcoffee for the inspiration for this.

volkflo:
Works good and LGTM after the changes we discussed :rocket:

volkflo (approved):

PR #1196, opened 2021-05-07
refactor(es): rework TLS configuration for Elasticsearch

### LINKED ISSUE #1195: Elasticsearch basic auth credentials are ignored when using https

volkflo:
Closes #1195

### DIFF
--- a/common/es/src/main/java/io/apiman/common/es/util/DefaultEsClientFactory.java
+++ b/common/es/src/main/java/io/apiman/common/es/util/DefaultEsClientFactory.java
@@ -19,8 +19,6 @@ import io.apiman.common.es.util.builder.index.EsIndexProperties;
 import io.apiman.common.logging.ApimanLoggerFactory;
 import io.apiman.common.logging.IApimanLogger;
 import io.apiman.common.util.Holder;
-import io.apiman.common.util.ssl.KeyStoreUtil;
-import io.apiman.common.util.ssl.KeyStoreUtil.Info;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -28,7 +26,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.KeyStore;
-import java.security.SecureRandom;
 import java.util.Date;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


common/es/src/main/java/io/apiman/common/es/util/DefaultEsClientFactory.java
```diff
@@ -198,12 +195,13 @@ private void waitForElasticsearch(RestHighLevelClient client, long pollingTime)
...
                             final Date endTime = new Date();
                             long pollingTimeMeasure = endTime.getTime() - startTime.getTime();
-                            logger.info("Took "+ pollingTimeMeasure + " milliseconds for polling Elasticsearch");
-
+                            logger.info("Took " + pollingTimeMeasure + " milliseconds for polling Elasticsearch");
                             // wake up the waiting thread
                             cdl.countDown();
                         }
-
+                    } catch (ElasticsearchException | SSLException e) {
```
msavy, line 202: Can you try just throwing these and see if it works?

And if not, I can tell you a few ways to handle this properly. For example, the holder can say whether it's a fatal error or not.

common/es/src/main/java/io/apiman/common/es/util/DefaultEsClientFactory.java
```diff
@@ -198,12 +195,13 @@ private void waitForElasticsearch(RestHighLevelClient client, long pollingTime)
...
+                        logger.error("Error connecting to Elasticsearch: ", e);
+                        exception.setValue(e);
                     } catch (IOException e) {
                         logger.info("Unable to reach Elasticsearch. Will continue polling.");
```
msavy, line 206: In the logger let's at least print out the exception message to guide us, let's say:

```
logger.info("Unable to reach Elasticsearch with error ({0}). Will continue polling.", e.getMessage())
```

volkflo: Will test everything, was just a draft pr to test if CI fails :)

volkflo:
I think this good to go for the moment.

- TLS is working now :heavy_check_mark:
- IO Exception Logging without annoying stack trace :heavy_check_mark:
- ES/TLS Exceptions with stack trace e.g. to find the wrong password :heavy_check_mark:

@msavy I think we could improve this e.g. that we stop the loop if the password is wrong but I think that is not that important at the moment.

msavy (approved):
I think this should be good to go now. TLS config fixes look good to me. Do we have this reflected in the example/sample configs and docs?

I've pulled out the connection polling code into a separate simple class that manages the process, which makes it easier to understand and doesn't need any hacks because all the code was enclosed within a method previously.

The other PR I opened has a fix for the Elasticsearch warning spam, so let's not replicate it here.

msavy:
BTW just in case you didn't notice, your commit message says 'Elasticse<b>ra</b>ch' instead of Elasticsearch 😅

PR #1205, opened 2021-05-12
refactor: create parent POM/BOM for whole Apiman platform

### LINKED ISSUE #989: Create a BOM (bill of materials) POM setup
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#bill-of-materials-bom-poms

A bill of materials (BOM, see link) will help us keep all the dependencies aligned between the 3 main Apiman repos. Apiman, Apiman Plugins, and Apiman Community Quickstarts.

At the moment we need to keep manually aligning all the the dependency versions which is error-prone and somewhat time consuming. 

A simple BOM will let us define all the versions centrally (you can still override as necessary)

msavy:
Closes #989


BastianGem (approved):
LGTM and builds fine on my machine :)
I am just a bit confused because no checks were started.

msavy:
It's likely excluded from the build rules

Issue #1082, opened 2021-03-24
Swagger and WDSL defs missing in manager exports


msavy:
API definition data is missing from the export-import definition.

PR #1242, opened 2021-05-26
fix(manager-api): add API Definitions to export-import

### LINKED ISSUE #1279: API definition data is missing from the export-import definition.

msavy:
API Definitions are missing from the export-import files. 

Fixes #1279


manager/api/export-import/src/main/java/io/apiman/manager/api/exportimport/manager/StorageExporter.java
```diff
@@ -97,33 +103,37 @@ public void export() {
...
-          Iterator<OrganizationBean> iter = storage.getAllOrganizations();
+            Iterator<OrganizationBean> iter = storage.getAllOrganizations();
+            int foo = 0;
 
-          while (iter.hasNext()) {
-              OrganizationBean bean = iter.next();
-              writer.startOrg(bean);
-              logger.info(Messages.i18n.format("StorageExporter.ExportingOrgs") + bean); //$NON-NLS-1$
+            while (iter.hasNext()) {
+                System.out.println("Organisation " + foo++);
```
msavy, line 120: Remove this println

msavy (commented):
Self-review OK, couple of nits to fix (stray diag statement)

volkflo:
Fixed https://github.com/apiman/apiman/issues/1082 as well :)

msavy:
Thanks a bunch! Couldn't find it; and apparently I wrote it 🤣.


>

msavy:
@volkflo @BastianGem I've changed the RestTest code a bit so that you need to provide the flag `X-RestTest-RegexMatching: true` to trigger regex matching on the response.

I realised this was causing some issues where it can unexpectedly trigger regex patterns (for example when brackets are used like this), which causes the tests to fail in a confusing way.

So now it's explicitly on (which makes it obvious that regex escaping is needed).

msavy (commented):
Self-review OK

volkflo (approved):
These changes look good to me. 
Import / Export may have a base64 bug.

volkflo:
Btw do you know if there is a default file upload limit on wildfly/tomcat?
The new export/import files are getting probably larger than before.

msavy:
> Btw do you know if there is a default file upload limit on wildfly/tomcat?
> The new export/import files are getting probably larger than before.

Good question. I think it can be configured in standalone.xml under  max-post-size.

msavy:
Looks like it's 10MB by default. Perhaps we should set it to something like 200MB?

msavy (commented):
self-review looks OK now

volkflo (approved):
LGTM now :)

PR #1243, opened 2021-05-26
build(parent): force junit4 as default for surefire plugin

msavy:
Ensure that TestNG does not barge in at any point transitively.

Also seems that junit5 dependencies declared in unnecessarily in certain places were causing some tests not to run.

I tried extensively to follow the instructions/examples from the surefire examples, but they didn't seem to work for me 🤷🏻‍♂️.

In future we should just move to JUnit 5.


volkflo (approved):
LGTM :)

Issue #1289, opened 2021-06-09
Example config files are not up to date


volkflo:
The example config files for vert.x, tomcat, and wildfly are not up to date.
The tomcat config file `apiman.properties` does not match the wildfly version.

PR #1290, opened 2021-06-09
refactor: harmonise config keys and refactor distro examples

volkflo:
Closes: #1289

This is related to the documentation PR https://github.com/apiman/apiman-docs-installation-guide/pull/4


msavy:
Let's figure out at some point this week or next how we can merge your ideas here in

volkflo:
@msavy What do you think about the idea of having settings that only contain one value configured in camel case?

Like we have for `allowSelfSigned` or `allowAnyHost`.
In the es settings, we currently have also `polling.time`, but `time` is the only setting in the block.

I would rather avoid such naming because this really makes the vert.x configs very unreadable because of such constructs:
```json
"polling": {
	"time": 600
}
```
What do you think about that idea?

distro/tomcat/src/main/resources/overlay/conf/apiman.properties
```diff
@@ -197,18 +235,21 @@ apiman-gateway.components.ISharedStateComponent.client.initialize=${apiman.es.in
...
-apiman-gateway.components.ISharedStateComponent.client.trust.host=${apiman.es.trust.host}
+apiman-gateway.components.ISharedStateComponent.client.allowSelfSigned=${apiman.es.allowSelfSigned}
+apiman-gateway.components.ISharedStateComponent.client.allowAnyHost=${apiman.es.allowAnyHost}
 
 
 # ---------------------------------------------------------------------
 # Rate Limiter Component Settings
 # ---------------------------------------------------------------------
 
+#apiman-gateway.components.IRateLimiterComponent=io.apiman.gateway.engine.impl.InMemoryRateLimiterComponent
```
msavy, line 251: I guess this should not be here? Or was this meant to be an example?

msavy (approved):

PR #1324, opened 2021-06-16
refactor(platform-vertx): improve error message when wrong buffer implementation used

msavy:


volkflo (approved):
Cool, looks reasonable to me :+1:

Issue #993, opened 2021-02-24
Having unique api gateway for each user


iamakashk:
Hi 
I am working with apiman for multi tenant application where in I want to have unique gateway for each customer or user of my application. I have gone through the documentation but could not figure out how manage gateway section is helping us.
![image](https://user-images.githubusercontent.com/19202603/108940941-3576d500-767a-11eb-89a1-ca16daec8b50.png)

I want to have following format:
For customer 1(Puma) my apiman gateway would look something like this.
[(http://localhost:8080/puma-gateway/path/to/resource](http://localhost:8080/puma-gateway/path/to/resource)

For customer 2(Nike) my apiman gateway would look something like this.
[http://localhost:8080/nike-gateway/path/to/resource](http://localhost:8080/nike-gateway/path/to/resource)
Is is possible to have unique gateway for each customer/user ? If yes how do we configure them?

Please let me know. Thanks for the help in advance!:)

volkflo:
Hi @iamakashk,
please have a look at this old issue: https://github.com/apiman/apiman/issues/711
Eric explained the gateway mechanism very well in the conversation. 

In short:
* You have to deploy multiple gateways (the easiest is the vert.x version)
* You have to configure a separate index/db-table for each gateway otherwise they will form a cluster

volkflo:
@msavy I think you referenced the wrong issue here?

msavy:
@iamakashk was @volkflo's answer sufficient for you? If so, can you close this issue.

If not, please feel free to expand :-).

iamakashk:
> @iamakashk was @volkflo's answer sufficient for you? If so, can you close this issue.
> 
> If not, please feel free to expand :-).

Thats really helped. Thanks @volkflo @msavy . As always you guys are saviors!!!😊😎

PR #1347, opened 2021-06-24
fix(manager-api): use correct configuration accessor for ESMetricsAccessor

### LINKED ISSUE #1346: Apiman Manager: can't use ES Metrics and non-ES store at the same time
Ok, so it looks like the initialisation code of ESMetricsAccessor was changed at some point so that it's now tightly bound to ES for the whole Manager configuration -- i.e. using ESMetricsAccessor will cause the system to try to look up configuration for the ES registry (`apiman-manager.storage.es` stuff). 

Naturally, this map will be empty because the user does not actually want to use the ES Registry for the Apiman Manager backend (i.e. e.g. using JPA/JDBC).

So, this needs to be disentangled before we can do the release.

I'll get it fixed tomorrow morning.

_Originally posted by @msavy in https://github.com/apiman/apiman/issues/933#issuecomment-867826158_

msavy:
`ESMetricsAccessor` should always use `config#getMetricsESClientFactoryConfig` *not*
`config#getStorageESClientFactoryConfig`.

Fixes #1346

### DIFF
--- a/common/config/src/main/java/io/apiman/common/config/options/exceptions/BadOptionConfigurationException.java
+++ b/common/config/src/main/java/io/apiman/common/config/options/exceptions/BadOptionConfigurationException.java
@@ -164,12 +164,7 @@ public class BadOptionConfigurationException extends IllegalArgumentException {
 
     @Override
     public String toString() {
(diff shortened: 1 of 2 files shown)


volkflo (approved):

msavy:
Should close #993

PR #1387, opened 2021-07-16
Initial Apiman migration assistant tool

msavy:
Backstory: We accidentally did not include the Api Schema definition (Swagger, etc) in the Apiman
export file. This means that it vanished when importing into a new version of Apiman. We can't fix
that with the existing tooling very easily because of complexities with Elasticsearch upgrades (the
schema isn't quite right).

So the simplest solution was to create this standalone tool which can export from the old version of
Apiman and then enrich the JSON with the missing information by invoking the Apiman Manager API.

The enriched file can then be imported into the new version of Apiman and the schemas will be
present.

Currently the tool just allows users to enrich pre-2.1.0.Final JSON to include API Schemas but in
future we may need it for other things.

It is presently is done in a very primitive way, and we can refactor it to do chaining of
patches or other stuff we might need if that need occurs.


migration/src/main/java/io/apiman/migration/fix_pre_21/ApimanExportVisitor.java
```diff
@@ -0,0 +1,93 @@
...
+
+                for (JsonNode versionNode : versionArray) {
+                    LOGGER.trace("Visiting API Version: {}", versionNode);
+                    visitApiVersion(orgId, apiId, versionNode);
+                }
+            }
+        }
+    }
+
+    public void visitOrg(String orgId, JsonNode orgNode) throws Exception {
```
volkflo, line 64: Whats the idea behind that?

volkflo (approved):
I think most of the changes are okay for now.
Would be maybe a bit nicer for the user to catch the exceptions and log something instead of getting the whole stack trace back (but that just a detail).

I tested it and it's working well!

msavy:
Yeah it's crude at the moment!

gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/connector/HttpConnector.java
```diff
@@ -231,11 +231,11 @@ public void transmit() {
...
 
-        if(clientRequest != null) {
-           clientRequest.end();
+        if (clientRequest != null) {
+            clientRequest.end();
         }
 
-        if(clientResponse != null) {
+        if (clientResponse != null) {
             clientResponse.netSocket().close(); //TODO verify
```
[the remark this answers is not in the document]
msavy: This is from years ago

migration/src/main/java/io/apiman/migration/fix_pre_21/ApimanExportVisitor.java
msavy: Idea behind what, sorry?

volkflo: Ahh sorry, was not clear enough.
I meant the empty void methods. Just preparation for the future or something like that?

msavy: The idea is that the person wanting to use this visitor can override those empty methods if they want in order to visit that particular entity. 

But it's optional, so they are not labelled as abstract methods but instead are just empty.

You can then do something like 

```java
new ApimanExportVisitor() {

  <snip>

  public void visitOrg(String orgId, JsonNode orgNode) throws Exception {
	System.out.println("Do something with " + orgId);
  }
};
```

volkflo: Okay, thank you for this example :) Makes sense now!

distro/wildfly/src/main/assembly/overlay.xml
```diff
@@ -100,6 +101,15 @@
         </includes>
       </unpackOptions>
     </dependencySet>
+<!--    <dependencySet>-->
```
volkflo, line 104: We do not need that?

volkflo (dismissed):

Issue #1374, opened 2021-07-08
Trust store in vert.x config assumes file exists 


msavy:
https://github.com/apiman/apiman/blob/94b22825e42a43266489db2246be68f0e1a89c17/distro/vertx/src/main/resources/overlay/configs/conf-es.json#L290

volkflo:
Maybe we should refactor the whole keycloak section at some point and also use jwks etc.

PR #1395, opened 2021-07-20
chore: comment out config samples with paths that may not exist

msavy:


msavy:
https://github.com/apiman/apiman/issues/1374

volkflo (approved):
Yep, makes sense :)

PR #1420, opened 2021-07-29
chore: set parent version in Apiman root pom to 2.1.1-SNAPSHOT

volkflo:


gateway/engine/beans/src/main/java/io/apiman/gateway/engine/beans/util/CaseInsensitiveStringMultiMap.java
```diff
@@ -456,5 +456,10 @@ public int getByte(String input, long offset) {
         public ByteOrder byteOrder(String input) {
             return ByteOrder.nativeOrder();
         }
+
+        @Override
+        protected Access<String> reverseAccess() {
+            return null;
```
msavy, line 462: Why have you left this null?

volkflo: Was just a test if it builds at all with adding that.

volkflo: Oh I just noticed that it is not a draft PR (I thought i pressed the button)

msavy (approved):

PR #1423, opened 2021-07-30
Refactor overlay files for tomcat and wildfly

volkflo:
The changed overlay files contain changes for the upcoming docker-compose quickstarts.

- Tomcat
  * Contains Keycloak Adapter in overlay
  * Contains web.xml/keycloak.json and context.xml for keycloak auth
- Wilfly
  * Added new apiman-slim-overlay for wildfly that does only contain the keycloak adapter
  * Removed Keycloak Server


volkflo:
Closed because of https://github.com/apiman/apiman/pull/1439 (so that @BastianGem can push to it)

manager/ui/war/tomcat8/src/main/webapp/META-INF/context.xml
```diff
@@ -0,0 +1,6 @@
+<Context path="/apimanui">
```
msavy, line 1: I wish we could do this Keycloak config as an external thing somehow instead of it being stuck inside of the WAR. I wonder if it's possible?

manager/ui/war/tomcat8/src/main/webapp/WEB-INF/keycloak.json
```diff
@@ -0,0 +1,23 @@
+{
```
msavy, line 1: And this too. I think it would be better if we could have keycloak.json outside of the WAR zip.

msavy: Regarding external `keycloak.json`: 

This doc seems to be for the OSGi adaptor, but I guess it should work for all of them; might require a bit of digging, but looks doable -- https://www.keycloak.org/docs/latest/securing_apps/#config_external_adapter

manager/ui/war/tomcat8/src/main/webapp/META-INF/context.xml
msavy: https://stackoverflow.com/questions/13956651/externalizing-tomcat-webapp-config-from-war-file

msavy:
Please reopen if still relevant.

PR #1569, opened 2021-11-11
fix(manager-api-jpa): include API version in query fetching API definition

### LINKED ISSUE #1567: Export problem: duplicate api definition
Please see the log for more details here: https://gist.github.com/volkflo/d17c3f4f447a108d29c69738004adf15

## Acceptance Criteria
- Return a maximum of one ApiDefinition for a given API Version.
- Export should not return any errors on export when multiple API Versions have schemas attached.

msavy:
The #getApiDefinition JPA query used in export was mistakenly missing the API version part of the
WHERE/AND predicate. This meant that if two different API versions had schemas attached, then the
query would return multiple entries.

Fixes #1567


msavy:
Cypress failure is expected. Waiting on contributor to provide a fix.

PR #1622, opened 2021-11-16
test(ui): fix sidebar interaction in cypress test 

BastianGem:
changed sidebar interaction from clicks to the corresponding mouse events
Additionally changed css selectors of sidebar elements to IDs

Cypress test works locally


msavy (approved):
Looks OK

At some point we need to add extra tests to cover all the new functionality and other stuff in a more granular way.

One idea I had is that we can use the import-export system to set up each scenario, and perhaps add a new function to quickly drop and reset the database (rather than having to reboot everything). Any thoughts on that?

msavy:
Thanks for your contribution!

Issue #1635, opened 2021-11-17
Add Notifications Link to Sidebar


BastianGem:
We should put the link to notifications page also in the sidebar 
There the link is once again more prominently placed and can not be overlooked

The link should also show the number of notifiactions without having to open the page.

PR #1636, opened 2021-11-17
feat(ui): add notifications to sidebar

BastianGem:
Link to notifications page also placed in sidebar
Number of notifications is displayed if greater than 0

notifications_sidebar

Closes: #1635

### DIFF
--- a/manager/ui/war/plugins/api-manager/html/headers/community.include
+++ b/manager/ui/war/plugins/api-manager/html/headers/community.include
@@ -194,6 +194,15 @@
                     </ul>
                 </div>
             </li>
+             <!--Notifications-->
+            <li class="list-group-item" data-target="#admin-secondary">
+                <a href="{{ pluginName }}/notifications">
+                    <span class="fa fa-bell" data-toggle="tooltip" title="Notifications"></span>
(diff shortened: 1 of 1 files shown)


manager/ui/war/plugins/api-manager/html/headers/community.include
```diff
@@ -194,6 +194,15 @@
                     </ul>
                 </div>
             </li>
+             <!--Notifications-->
+            <li class="list-group-item" data-target="#admin-secondary">
```
volkflo, line 198: I think we should add an ID here for the cypress tests? What do you think?

volkflo (changes requested):

msavy:
For the community version, do we need this? It's already there on the top right and this is just duplicating it, no?

msavy:
@EricWittmann any super-fast impressions/thoughts?

For Apiman community should we:
- Have 'notifications' link + count on both the sidebar and header bar (see screenshot: top right of header, and last entry of side menu)
- Only have it on the header bar
- Only have it on the sidebar

Given both of them track you around on every page, I'm not sure it make sense to have both -- but I don't mind it either 🤔

Issue #1625, opened 2021-11-16
Page truncated after clicking save on new definition for Developer Portal


kgebert:
**Setup**
- Apiman version 2.1.1 final (I guess as this is in PAS 21.2)
- Environment
  - Linux
  - pas-doc

**Bug**
After having saved a new definition for a Developer Portal within an API the page is truncated at the top:

![image](https://user-images.githubusercontent.com/28644593/142006075-e7388af0-f92e-49c4-bdf2-9a0cec92d8d2.png)

User cannot access the breadcrumbs anymore.
The page stays truncated even if I switch to a different page. This is only when I trigger a reload via the browser.

There's nothing special in the Console:
`PUT https://scheer-acme.com/pas-doc/apiman/organizations/PRODUCTION/apis/SAP_Fulfillment/versions/1.1`
`apiman [3:47:09 PM]>>  Api Version update succeeded!`

**To Reproduce**

1. Go to an **API details** page.
2. Go to section **Developer Portal**.
3. Set the API up or change something.
4. Click **Save**.

==> Page is truncated at the top.

msavy:
@volkflo  this doesn't happen in upstream. Are you OK with me closing this?

volkflo:
It happens in a different shape on upstream.
@BastianGem tested this and in upstream we have two scrollbars. If you remove one, we have the same problem.
Please see on the right side:
![image](https://user-images.githubusercontent.com/13332383/142204980-06765924-5013-49f6-a8ae-df9cedca8883.png)

msavy:
Interesting. I don't get that, but might be because my OS doesn't show bars 😂.

volkflo:
That issue seems to be related with this one: https://github.com/nhn/tui.editor/issues/1802
If your are not at buttom of the page and do a reload, the editor gets the focus and tries to scroll down.
After that we have this problem.

volkflo:
@msavy I updated the version of the editor because they fixed the autofocus thing.
But there are still some problems in the editor itself that could cause weird behavior in markdown mode.

- https://github.com/nhn/tui.editor/issues/1777
- https://github.com/nhn/toast-ui.vue-editor/pull/37

So I switched the initial mode to wysiwyg (which so be okay for most of the people).
With that setting the page stops scrolling down to the editor and does not get the focus automatically.

volkflo:
Fixed in 3.x

PR #1638, opened 2021-11-18
fix: change behaviour of markdown editor and fix css issues

volkflo:
Closes: #1625


volkflo:
Still not completely fixed but its a first draft

volkflo:
I think we should keep the issue open till bug in tui editor is addressed.

msavy (approved):

volkflo:
Merge that temporarily and try to fix that at some later point to start with markdown mode per default

PR #1725, opened 2021-12-22
refactor(distro-wildfly): align Apiman to Keycloak 15.1.1, which required upgrading to WildFly 23.0.2

msavy:
refactor: Align Apiman with  Keycloak 15.1.1, which required upgrading to WildFly 23.0.2

This mostly included reworking the WildFly quickstart distro's `standalone-apiman.xml` config (again).

Now using a self-signed cert instead of default certificate. Could there be unintended consequences to this? For example, it is not explicitly linked to the apiman keystore, so could that mess with existing configs.

### DIFF
--- a/distro/wildfly/pom.xml
+++ b/distro/wildfly/pom.xml
@@ -13,7 +13,7 @@
 
   <properties>
     <docker.wildfly.version>wildfly</docker.wildfly.version>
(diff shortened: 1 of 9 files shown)


msavy:
I'm not sure it's a good idea to change to self-signed certs yet because people are probably relying on apiman.jks being linked up...

We should do one more release of 2.x with this old-style I guess and then switch over (clear migration guide notes).

PR #1755, opened 2022-01-19
refactor(manager-api-rest): use DTO for old devportal #getAllApiVersions

msavy:
To avoid Jackson IdentityInfo references leaking to presentation layer we add a simple DTO.


volkflo (approved):
Looks like a good solution for now!
Can you merge it when you are ready? :)

PR #1775, opened 2022-01-27
fix(ui): add local time to time-restricted-access-policy

volkflo:
- fixing DateTime issues that where caused by previous library replacement of moment
- adding local time along utc time view in table
- changing some functions to anonymous functions to avoid unexpected behaviour

@msavy The same PR you already reviewed :)


msavy:
Rebased

msavy (approved):

PR #1785, opened 2022-02-04
fix(ui): modified status label restored by removing self-closing span

volkflo:
valid html5 requires closing tags on span elements


volkflo:
The failed cypress test is not related to this fix, its related to the exception we discussed yesterday @msavy

PR #1842, opened 2022-03-01
feat(manager-api): if mutable user attributes change in IDM, synchronise user's Apiman profile

msavy:


msavy:
@volkflo would appreciate if you can try this out before I merge in 👍. Seems to work as intended.

volkflo (approved):
LGTM, as we looked on it together.
Maybe add the ticket number to you TODO comment before merge (#1849) to find that one easily

PR #1843, opened 2022-03-02
Correct system endpoint in ApimanManagerConfig

msavy:


volkflo (approved):
LGTM :) Tested it locally, works fine!

PR #1847, opened 2022-03-02
Add user notification preferences relationship to UserBean

msavy:


msavy:
@volkflo would appreciate if you can try this out against your setup and verify -- all seems good, but double check would be great!

msavy:
Please note this tiny change to DDL to add FK. Suggest you just run it manually 👍. It'll probably still work without it, but this is more optimal.

https://github.com/apiman/apiman/pull/1847/files#diff-a91051ba86ddd12f4fc42ecf37172d3edde16486da8aefe2f146a189fb733a91R419-R421

Issue #1884, opened 2022-03-29
ES Metrics: if metrics buffer is full, then drop records rather than blocking


msavy:
If ES goes down the buffer will eventually fill up, and the calling thread will be blocked by the BlockingQueue. Although this is rare, it can cause the thread to be blocked forever.

Instead we should drop the records and print an error into the logs (possibly including the missing metrics).

This is particularly impactful for the Vert.x Gateway where we should never block the thread.

The underlying causes could be:
- ES is down and does not recover (at all, or after a long time)
- Networking problems where ES is not reachable
- Client threads end up stuck permanently, so metrics are eventually never sent (see some comments in https://github.com/elastic/elasticsearch/issues/65213 for similar issues)
- Very high rate of metrics accumulation but unable to write them fast enough (most likely due to slow ES server)

msavy:
See https://github.com/elastic/elasticsearch/issues/65213 (newer comments towards bottom of thread) for other people with similar issues.

PR #1886, opened 2022-03-29
feat(metrics): drop new RecordMetrics in ESMetrics if buffer is full, rather than blocking

msavy:
If ES goes down the buffer will eventually fill up, and the calling thread will be blocked by the
BlockingQueue. Although this is rare, it can cause the caller to be blocked forever.

This is a particular problem for the Vert.x gateway where it is not acceptable for calling threads
to be blocked for long periods of time.

Fixes: #1884


volkflo (approved):
Great, see comment in #1885

Issue #1887, opened 2022-03-30
Policies can be left dangling when deleting


BastianGem:
**Describe the bug**
If you delete a plan / client / API with an attached policy or an organization where a plan / client / API with an attached policy resides in , the policy itself won't be deleted.

Steps to reproduce the behaviour:
1. Create an organization and a plan, client or API.
2. Attach a policy to the plan, client or API.
3. Delete the plan, client, API or the whole organization.
4. Recreate the plan, client, API or the whole organization (and the plan, client or API in it). While recreating you'll have to choose the exact same name.
5. Find the still configured policy in your plan, client or API

**Expected behaviour**
If you delete a plan, client, API or the whole organization, everything inside should be deleted as well.

PR #1888, opened 2022-03-30
fix(manager-api-jpa): ensure policies are not left dangling after deletion

BastianGem:
If a client / plan /api gets deleted, all policies attached to it should be deleted as well.
To fix the current behaviour that the policies remain in the database after the deletion of a plan / api / client, a missing executeUpdate is added to the deleteAllPolicies function
To ensure future changes won't break this logic a new delete test is added as well

Closes: #1887

### DIFF
--- a/manager/test/api/src/test/java/io/apiman/manager/test/DeleteTest.java
+++ b/manager/test/api/src/test/java/io/apiman/manager/test/DeleteTest.java
@@ -1,0 +1,16 @@
+package io.apiman.manager.test;
+
+import io.apiman.manager.test.junit.ManagerRestTestPlan;
+import io.apiman.manager.test.junit.ManagerRestTester;
+
+import org.junit.runner.RunWith;
+
+/**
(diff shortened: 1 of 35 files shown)


manager/test/api/src/test/java/io/apiman/manager/test/DeleteTest.java
```diff
@@ -0,0 +1,16 @@
...
+
+import io.apiman.manager.test.junit.ManagerRestTestPlan;
+import io.apiman.manager.test.junit.ManagerRestTester;
+
+import org.junit.runner.RunWith;
+
+/**
+ * Runs the "delete" test plan.
+ *
+ * @author eric.wittmann@redhat.com
```
msavy, line 11: I think you did this, Bastian 😅

msavy (approved):

msavy:
I'll backport this onto 2.x also. Thanks for this contribution, @BastianGem -- much appreciated.

PR #1899, opened 2022-04-08
fix(manager-api-jpa): ensure policies are not left dangling after deletion

msavy:
If a client / plan / API gets deleted, all attached policies should be deleted as well.

To fix the current behaviour,  where the policies remain in the database after the deletion
of a plan / API / client, a missing `executeUpdate` is added to the deleteAllPolicies function.

To ensure future changes won't break this logic a new delete test is added, as well.

Closes: #1887


msavy:
Backport onto master. Please have a look.

Issue #1451, opened 2021-08-17
[Gateways][2.1.0] Unable to set different index metrics name for each gateway instance


OSMOSA44:
Bonjour,

In our configuration, we have decided to create elasticsearch indices per gateway (more than 20 gateways)

With version 1.5.1, there was no problem with this configuration fixing `index` propertie for each component (snippet of relevant configuration below)

```xml
apiman.es.index=${env:APIMAN_GATEWAY_REGISTRY_CLIENT_INDEX}
apiman-gateway.registry.client.index=${apiman.es.index}_registry
apiman-gateway.metrics.client.index=${apiman.es.index}_metrics
apiman-gateway.components.ISharedStateComponent.client.index=${apiman.es.index}_share
apiman-gateway.components.IRateLimiterComponent.client.index=${apiman.es.index}_quota
apiman-gateway.components.ICacheStoreComponent.client.index=${apiman.es.index}_cache

```

With version 2.1.0, we want to keep the same configuration but `index` has been replaced by `indexPrefix` and the behaviour differs. With the configuration below

```bash
apiman.es.index=${env:APIMAN_GATEWAY_REGISTRY_CLIENT_INDEX}
apiman-gateway.registry.client.indexPrefix=${apiman.es.index}_registry
apiman-gateway.metrics.client.indexPrefix=${apiman.es.index}_metrics
apiman-gateway.components.ISharedStateComponent.client.indexPrefix=${apiman.es.index}_share
apiman-gateway.components.IRateLimiterComponent.client.indexPrefix=${apiman.es.index}_quota
apiman-gateway.components.ICacheStoreComponent.client.indexPrefix=${apiman.es.index}_cache
```

The first launch give us on elasticsearch those indices (with env APIMAN_GATEWAY_REGISTRY_CLIENT_INDEX=gw_test)

```bash
yellow open gw_test_metrics              ZnqEP2s0T5izK3lTY9drzg 1 1 0 0 208b 208b
yellow open gw_test_registry_apis        2AEQK4r7QVWJSYw7JE8zxw 1 1 0 0 208b 208b
yellow open gw_test_registry_clients     uNAjbeFxSOmTsPZdADmtVw 1 1 0 0 208b 208b
yellow open gw_test_registry_dataversion 9jpMxnsZQGWLNABGUvUPuw 1 1 0 0 208b 208b
```

On next step i declare a test api and request it, that generates a new `apiman_metrics` indice which will be global for all my gateway instances

```bash
[...]

PR #1940, opened 2022-04-20
fix(metrics-es): use custom index prefix without unnecessary underscore

msavy:
Add backwards compatibility with `client.index` key alias.

Fixes: #1451

### DIFF
--- a/manager/api/es/src/main/java/io/apiman/manager/api/es/EsStorage.java
+++ b/manager/api/es/src/main/java/io/apiman/manager/api/es/EsStorage.java
@@ -2005,7 +2005,7 @@ public class EsStorage extends AbstractEsComponent implements IStorage, IStorage
     }
 
     private String getFullIndexName(String indexPostFix) {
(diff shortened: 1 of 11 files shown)


manager/api/es/src/main/java/io/apiman/manager/api/es/EsStorage.java
```diff
@@ -2007,7 +2005,7 @@ private void indexEntity(String type, String id, XContentBuilder sourceEntity)
     }
 
     private String getFullIndexName(String indexPostFix) {
-        return (getIndexPrefix() + "_" + indexPostFix).toLowerCase();
+        return (getIndexPrefixWithJoiner() + "_" + indexPostFix).toLowerCase();
```
msavy, line 2008: looks like this would cause 2 underscores with previous approach?

msavy:
Ping: @OSMOSA44 @shachindrasingh

PR #1943, opened 2022-04-22
refactor(distro): remove Keycloak Server from WildFly distro

msavy:
- [x] Remove Keycloak server overlay from Apiman distros.
- [x] Create Keycloak docker-compose setup in server-all for localhost development/demo purposes. This ensures we still have a convenient development environment. 
	- [x] Pull Keycloak version from Apiman BOM/parent.
	- [x] Use a sensible default KC definition file.
- [x] Ensure devserver still works 
	- [x] Set environment/run configs to ensure docker-compose file runs before devserver
	- [x] Shutdown, stop compose server
- [x] Bump KC version, where possible.

Does the new AIO docker-compose distro live in main Apiman repo or docker repo?


msavy:
Can't bump to newest version of WildFly as Keycloak OAuth adapter doesn't work any more, and the replacement generic OIDC adaptor doesn't support bearer-only auth.

msavy:
@volkflo this change might interest you

Issue #1181, opened 2021-05-05
Browser back button does not work on "All APIs" page


volkflo:
After having switched to "All APIs", user cannot go back to the previous page by hitting the browser back button. He will stay on the "All APIs" page.

PR #2005, opened 2022-05-05
fix(ui): browser back button on "all"-pages

volkflo:
By directly linking to the URL with query parameters we automatically load the first page. This avoids reloading the first page via `loadFirstPage()` in `consumer-orgs/apis.ts` which would create two browser history entries.
Keeping `loadFirstPage()` is still useful to display the first results after reloading the page or after resetting the search.

Closes: #1181

### DIFF
--- a/manager/ui/war/plugins/api-manager/html/headers/community.include
+++ b/manager/ui/war/plugins/api-manager/html/headers/community.include
@@ -71,7 +71,7 @@
                             </a>
                         </li>
                         <li class="list-group-item" id="apiman-sidebar-apis-consume" data-target="#apis-all_apis-tertiary">
-                            <a href="{{ pluginName }}/browse/apis">
+                            <a href="{{ pluginName }}/browse/apis?q=*&cp=1&ps=12">
                                 <span class="list-group-item-value" apiman-i18n-key="actions.find-api">All APIs</span>
                             </a>
                         </li>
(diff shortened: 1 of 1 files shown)


msavy (approved):

manager/ui/war/plugins/api-manager/html/headers/community.include
```diff
@@ -71,7 +71,7 @@
                             </a>
                         </li>
                         <li class="list-group-item" id="apiman-sidebar-apis-consume" data-target="#apis-all_apis-tertiary">
-                            <a href="{{ pluginName }}/browse/apis">
+                            <a href="{{ pluginName }}/browse/apis?q=*&cp=1&ps=12">
```
msavy, line 74: Can you explain the `cp` and `ps`?

volkflo:
Sorry forgot one file will push a fixup in a minute

manager/ui/war/plugins/api-manager/html/headers/community.include
msavy: Current page and page size I guess

volkflo: Ah yes, so the browse/consume page uses the search endpoint.
So to avoid displaying an empty page on first page load we fetch the first page of a wildcard search (SearchBean).

```bash
q = querry
cp = currentPage
ps = pageSize 
```
This is a nice default which does not display to much data and plays nice with our 4x3 card layout.

volkflo: image

msavy: Did you guys add that?/ I don't remember that being the behaviour Apiman had historically

volkflo: I think that is the case since a while, apiman 2.x behaves the same.
You can change the page size in the UI if there are more entries to 25, 50 and 100 I think

msavy: Okay, let's just leave it for now. Perhaps needs to be a bit more flexible and signpost that you can search more easily.

msavy (approved):

PR #2027, opened 2022-05-10
feat(policies): blacklist/whitelist add support for IPv6, CIDR, ranges.

msavy:


msavy:
https://user-images.githubusercontent.com/423513/167944987-11c939de-31e0-475d-bed1-0325645c3157.mov

msavy:
BTW, this has been implemented in a way that _should_ be fully backwards compatible (the parser on the backend infers whether it's IPv6 vs IPv4, there's some simple logic to pick out ranges, etc.). 

So, please test out any existing rulesets you have.

volkflo:
@msavy There might be a problem with the cache as the cache entries are never invalidated.

Currently if I make a request that should pass the IP is added into the cache as (IP, false) which is okay.
Later I decide to add a rule to the policy that should block this IP but the IP is not blocked because the cache still contains the old entry.
After I restart the gateway the cache is clean again and the IP is added as (IP, true).

msavy:
> @msavy There might be a problem with the cache as the cache entries are never invalidated.
> 
> Currently if I make a request that should pass the IP is added into the cache as (IP, false) which is okay. Later I decide to add a rule to the policy that should block this IP but the IP is not blocked because the cache still contains the old entry. After I restart the gateway the cache is clean again and the IP is added as (IP, true).

Thanks. I will double check my logic for the cache keys, I was meant to include the configuration in the key (hence adding a hash function to the config!), but looks like I forgot about it at some point.

msavy:
@volkflo please try again 👍

volkflo:
Okay tested it again, as far as I can tell it works fine now :+1: 
These changes seem like a really nice addition to the policy!

manager/ui/war/plugins/api-manager/html/policyForms/ip-list.include
```diff
@@ -1,33 +1,48 @@
...
     </div>
   </div>
-  <input id="ip-address" ng-model="ipAddress" ng-pattern="/((^\s*(((?:\*|([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]))\.){3}(?:\*|([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])))\s*$))/" data-field="ipAddress" class="form-control" style="width: 200px; float: left; margin-right: 5px" type="text" apiman-i18n-key="iplist.enter-ip-address" placeholder="Enter an IP address..." ng-disabled="isEntityDisabled()">
-  <button id="add" ng-disabled="!ipAddress || isEntityDisabled()" ng-click="add(ipAddress)" data-field="add" apiman-i18n-key="add" class="btn btn-default" style="min-width: 75px">Add</button>
-</div>
+  <!-- Use ng-model-options allowInvalid, otherwise value is not set until it passes validation, meaning we can't display any useful hints/prompts for user -->
+  <input id="ip-address" name="ipinput" ng-model="ipAddress" data-field="ipAddress" class="form-control" style="max-width: 300px; float: left; margin-right: 5px" type="text" apiman-i18n-key="iplist.enter-ip-address" placeholder="Enter an IP address..." ng-disabled="isEntityDisabled()" ng-model-options="{ updateOn: 'default', allowInvalid:'true' }" validate-ip-address/>
```
volkflo, line 43: @msavy Can we try to avoid these inline styles, as they are really annoying to overwrite if someone wants to do a little bit of customization

volkflo (approved):

manager/ui/war/plugins/api-manager/html/policyForms/ip-list.include
msavy: I'm going to completely re-format the TS and HTML of the project once we've done 3.x (using prettier or similar). 

I agree having very long lines like this makes it difficult to spot the difference (I only added `name` and changed width slightly.

msavy:
Thanks for testing and feedback, will merge today.

Issue #2030, opened 2022-05-11
Delete button for locked policy shown but leads to error


volkflo:
**Problem**

Currently the "Remove" button appears on a policy even if the plan is in locked state.
Removing a policy of a locked plan, results in an error message:
"Already Exists You tried to create something that already exists. Please try again with a different name"

**Acceptance**

A locked plan cannot be altered, therefore the delete button for a policy should also not be visible.

volkflo:
Okay, it seems this also a problem on all pages that are dealing with policies (clients, apis).
The button is also enabled for a retired API.

PR #2031, opened 2022-05-11
fix(ui): only show remove button if removal is possible

volkflo:
A locked plan cannot be altered, therefore the delete button for a policy should also not be visible. This also applies for retired APIs.
Fixing the problem by passing the entity check from parent to child component.

Closes: #2030


volkflo:
I am not sure about the failing java 17 test. Seems that this is not related.

volkflo:
> Reminder to self, let's discuss this one soon, as I'm not quite sure how it works and how it would affect other components.

Sure just give me a ping

msavy (approved):
Let's double check this, but I _think_ it's OK

Issue #2033, opened 2022-05-11
Layout in 'Edit' versions of policies has large left padding/margin


msavy:
![image](https://user-images.githubusercontent.com/423513/167945759-0654c5e8-1638-4480-acfe-4652b5149c7e.png)

This is not present when the policy is first inserted (add-policy).

Looks like this was introduced recently?

Should not be too complicated to fix.

volkflo:
I will have a look, I think I know where this comes from.

PR #2034, opened 2022-05-12
fix(ui): fix display of policy edit view

volkflo:
Remove css classes to be consistent with "new-policy" view.

Closes: #2033


msavy (approved):

msavy:
Failure is a problem with Maven Central. Will kick it again later.

Issue #873, opened 2020-07-28
Rename Blacklist/Whitelist policies


kgebert:
In respect to the recent discussions all over the world, it would be nice to rename the mentioned policies to Allowlist/Denylist.

EricWittmann:
Agreed.  Thank you for opening this.

msavy:
We should have a think about how we could effect this change without accidentally breaking backwards compatibility. And how deep we should go with the incisions.

Initial thoughts:

First phase might just be to change the UI labelling as that should be (almost?) completely safe. That seems low hanging fruit and entirely doable.

However, a full migration in code might actually be impossible without causing serious incompatibility/breakages. 

I've been doing an audit of the relevant code, and there are loads of places that reference the relevant policies by their (internal) coordinates and names, including in persistent stores (i.e. a name change would break backwards compatibility, as a change to the name would change the name of the policy's storage area making it seem 'empty').

Thoughts?

EricWittmann:
I agree that it's a non-trivial change and will cause backwards compatibility issues.  +1 to immediately changing the UI labeling.  That's a no-brainer.  After that we'll have to play with it.

volkflo:
Just another idea, but I think not the perfect one:
I looked a bit into the both policies and they both missing two important features.
- no support for IPv6
- no support for proxy headers like X-Forwarded-For (it is a list of IP addresses and the policy can not handle a list of IP address from a header)

So maybe it would be nice to create two new policies with a different name and more features and deprecate the old ones.

EricWittmann:
I think that could be a nice compromise - where we do two things:

1) Rename the existing policies (UI labels only)
2) Deprecate existing policies and replace with new, better versions

That's probably better and easier than trying to figure out all the touch points of a proper/full rename.

PR #2040, opened 2022-05-13
refactor(distro-data): change policy names blacklist -> blocklist; whitelist -> allowlist

msavy:
Fixes: #873


EricWittmann:
nice

Issue #2046, opened 2022-05-17
Wrong permission check on approve contract button


BastianGem:
**Describe the bug**
Due to a wrong permission check on the approve contract button, it is currently only possible to approve a contract if you have the IDM role 'apiadmin'.

**To Reproduce**
1. Create an APIwith an attached Plan which requires approval and expose it to the devportal
2. Sign Up to the API with a devportal user.
3. Try to approve the new contract with an apiuser with the Apiman permission planAdmin
4. The approve button is missing

**Expected behaviour**
Every user with the permission planAdmin in this organization should be able to approve new contracts.
In addition every user with the IDM role apiadmin should also be allowed to approve new contracts

PR #2047, opened 2022-05-17
fix(ui): fix permission check of contract approve button

BastianGem:
Due to a wrong permission check on the approve contract button only user with the IDM role 'apiadmin' were able to approve contracts.

We fixed this permission check, that it only checks the apiman permission 'planAdmin' now. User with the IDM role 'apiadmin' will still bypass this permission check.

Closes: #2046


msavy (approved):
Interesting. Thanks for this. Did the comma mean you had to be both? 🤔

I thought it was meant to be `planAdmin,admin` meant `planAdmin or admin`.

Agreed `admin` not needed as that bypasses the check anyway.

Thanks for this!

Issue #2049, opened 2022-05-19
Image Upload doesnt work anymore


BastianGem:
**Describe the bug**
When trying to upload an image for an API (Devportal) you always get an error saying 'You must be logged in'.
It doesnt matter if you're logged in or not, the behaviour is always the same.

**To Reproduce**
1. Create an API
2. Go to Devportal Tab
3. Click on 'Upload Image'
4. Choose an image
5. Click on Upload
6. See Error

**Expected behaviour**
This Error  should only be shown if you're not logged in.

PR #2050, opened 2022-05-19
fix(manager-api): fix image upload

BastianGem:
Because of a missing entry in the web.xml (wildfly and tomcat) the loggedIn-check for uploading images didnt work properly.
We added the entry for the /blobs endpoint back in, to fix the loggedIn-check.

Closes: #2049


msavy (approved):
Thanks, Bastian! I appreciate this.

msavy (approved):

Issue #2069, opened 2022-05-25
Without IDM apiadmin role no users can be added to own organization


volkflo:
2 users max.mustermann & willy.welcome have been removed from the IDM apiadmin role.

Willy Welcome now wants to add max.mustermann  to his own organization and selects the role Editor for him - as soon as he clicks on "Add Member(s)" he gets an error message that he does not have sufficient permissions. 
![image](https://user-images.githubusercontent.com/13332383/170232412-cf37948b-9a16-4428-87dc-15f2aea83678.png)

max.mustermann  is also not added to the organization. 
If willy.welcome rejoins the group API Admins - he can invite without problems. 

## Acceptance criteria
- Non super-admin user must be allowed to invite people into orgs (must have `orgAdmin` role).

PR #2070, opened 2022-05-25
fix(manager-api): grant membership as non-admin user

volkflo:
By changing from `IUserResource` to the `UserService` directly we avoid the `checkIfUserIsCurrentUser` in the security context.
This check was a problem during granting memberships as non-admin user because it was not allowed to get the user object as non-admin / as non-being the user itself.

Closes: #2069


volkflo:
This need backporting to 2.x

msavy (approved):

msavy (approved):

PR #2120, opened 2022-06-21
fix(ui): print error message in modal

volkflo:
@msavy You already reviewed this one in our downstream fork.

### DIFF
--- a/manager/ui/war/plugins/api-manager/ts/modals.ts
+++ b/manager/ui/war/plugins/api-manager/ts/modals.ts
@@ -82,11 +82,13 @@ _module.factory('Modals',
                     try {
                         switch (typeof(rpcdata)) {
                             case "string":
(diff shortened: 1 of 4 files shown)


manager/ui/war/plugins/api-manager/ts/modals.ts
```diff
@@ -82,11 +82,13 @@ _module.factory('Modals',
                     try {
                         switch (typeof(rpcdata)) {
                             case "string":
-                                message = JSON.parse(rpcdata).message;
```
msavy, line 85: Nice. The `??` feature wasn't in TS when we kicked off Apiman, IIRC.

msavy (approved):

msavy:
Thanks for your contribution @volkflo -- as always, it's appreciated.

PR #2121, opened 2022-06-21
fix(ui): remove warning in developer portal tab

volkflo:
@msavy You already reviewed that one in our downstream fork


msavy (approved):

msavy:
Thanks @volkflo!

PR #2131, opened 2022-06-24
docs(migration): document initial 3.x breaking changes

msavy:
* Elasticsearch removal

* WildFly + Keycloak AIO distro removal

* Docker AIO removal


msavy:
We should list docker images, link docker-compose, etc, once we're at that stage.

PR #2190, opened 2022-07-15
feat(notifications): add email to user on api signup

volkflo:
Add a new notification that is sent to the user who signs up to an API. This is sent in addition to the approval notifications to all plan admins.

Closes: #1852


msavy (changes requested):
Can you add tests in `DefaultTemplatesTest`. I just realise the previous contribution also didn't have a test, so one needs adding there, too.

This ensures the events actually work, otherwise we sometimes won't find out the template is defective until someone tries to run it IRL.

volkflo:
> Can you add tests in `DefaultTemplatesTest`. I just realise the previous contribution also didn't have a test, so one needs adding there, too.
> 
> This ensures the events actually work, otherwise we sometimes won't find out the template is defective until someone tries to run it IRL.

Totally forgot about that, sorry.
Could directly catch a little reconfiguration in the new templates. :+1:

msavy (approved):

Issue #2209, opened 2022-07-27
Access denied when creating API contract with certain visibilities combinations


msavy:
### Apiman Version

3.0.0-SNAPSHOT

### Apiman Manager Distro

Tomcat

### Apiman Gateway Distro

Vert.x

### Java Version

❯ java -version
openjdk version "11.0.15" 2022-04-19
OpenJDK Runtime Environment Temurin-11.0.15+10 (build 11.0.15+10)
OpenJDK 64-Bit Server VM Temurin-11.0.15+10 (build 11.0.15+10, mixed mode)

### Operating System

macOS Monterey 12.5

### Are you running Apiman in a container, or on an orchestration platform?

Docker Compose

### Describe the bug

Original reporter is @volkflo ... I've augmented the bug report with a fair amount of additional information I discovered during initial investigation process

---

Visibility of plans is not being properly indexed in the case of multiple plans for same API which have certain configurations of different/conflicting discoverability levels.

The data is correct in the DB:

![image](https://user-images.githubusercontent.com/423513/181380652-650a8f4e-8440-4201-b346-aa88f7f48710.png)

But I think there is something missing/incorrect here as there is no "Petstore.PLAN.Basic.VERSION.1.0 for the Petstore API only one entry for the  "Petstore:Test:1.0:Basic:1.0" API

![image](https://user-images.githubusercontent.com/423513/181380667-f9e78f64-2465-4916-a3c6-959b01a130e7.png)

### Expected behaviour

The plan:
- Org: Petstore
- PlanId: Basic 

Should be visible to non-members, by virtue of visibility of Petstore:Basic version `1.0` having `PORTAL` visibility.

### Actual behaviour

Index erroneously returns non-discoverable and hence causes forbidden.

### How to Reproduce

Dataset to reproduce

```csv
id,org_id,api_id,api_version,plan_id,plan_version,discoverability
Petstore:APIWithoutPlans:1.0,Petstore,APIWithoutPlans,1.0,,,ORG_MEMBERS
Petstore:Metrics:1.0,Petstore,Metrics,1.0,,,ORG_MEMBERS
Petstore:Metrics:1.0:Sandbox:1.0,Petstore,Metrics,1.0,Sandbox,1.0,ORG_MEMBERS
Petstore:Metrics:1.0:Gold:1.0,Petstore,Metrics,1.0,Gold,1.0,ORG_MEMBERS
[...]

PR #2211, opened 2022-07-28
fix(manager-api-security): plans added to discoverability index must inherit highest discoverability

msavy:
Fixes: #2209


volkflo (approved):
Tested with json that was attached to the issue #2209.
Fix works as expected.
I like that you added some comments in `insertIntoIndex` :+1:

msavy:
> Tested with json that was attached to the issue #2209. Fix works as expected. I like that you added some comments in `insertIntoIndex` 👍

Thanks so much for the verification!

Issue #2213, opened 2022-07-28
Blank "definitonUrl" is not handled well and lead to unnecessary execption


volkflo:
### Apiman Version

3.0.0-SNAPSHOT

### Apiman Manager Distro

Tomcat

### Apiman Gateway Distro

Vert.x

### Java Version

11

### Operating System

Linux

### Are you running Apiman in a container, or on an orchestration platform?

Kubernetes

### Describe the bug

Empty string in `defintionUrl`while creating a new API leads to exception:
```java
[ERROR] 2022-07-23 07:35:55.500 [http-nio-8080-exec-2] ApiService - Unable to store API definition from: 
java.net.MalformedURLException: no protocol: 
    at java.net.URL.<init>(URL.java:645) ~[?:?]
    at java.net.URL.<init>(URL.java:541) ~[?:?]
    at java.net.URL.<init>(URL.java:488) ~[?:?]
    at io.apiman.manager.api.rest.impl.util.ScheerPasBridgeConnection.loadDefinitionFromUrl(ScheerPasBridgeConnection.java:59) ~[apiman-manager-api-rest-impl-8.3.0-SNAPSHOT.jar:?]
    at io.apiman.manager.api.service.ApiService.createApiVersionInternal(ApiService.java:492) ~[apiman-manager-api-rest-impl-8.3.0-SNAPSHOT.jar:?]
    at io.apiman.manager.api.service.ApiService$Proxy$_$$_WeldSubclass.createApiVersionInternal(Unknown Source) ~[apiman-manager-api-rest-impl-8.3.0-SNAPSHOT.jar:?]
    at io.apiman.manager.api.service.ApiService.lambda$createApi$3(ApiService.java:241) ~[apiman-manager-api-rest-impl-8.3.0-SNAPSHOT.jar:?]
    at io.apiman.manager.api.rest.impl.util.DataAccessUtilMixin.tryAction(DataAccessUtilMixin.java:16) ~[apiman-manager-api-rest-impl-8.3.0-SNAPSHOT.jar:?]
    at io.apiman.manager.api.service.ApiService$Proxy$_$$_WeldSubclass.tryAction$$super(Unknown Source) ~[apiman-manager-api-rest-impl-8.3.0-SNAPSHOT.jar:?]
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
[...]

PR #2214, opened 2022-07-28
fix(manager-api): add isBlank check to `definitionUrl`

volkflo:
Avoid unnecessary `MalformedURLException`s by checking the string first.

Closes: #2213


msavy (dismissed):

msavy:
Looks good to me. Just wonder if we should have a test, but just thinking where the best place might be.

msavy (approved):

volkflo:
Mhh agree, we should think about getting some unit tests for the services at some point. Not everything needs a full `.resttest`.
Not sure how good that works currently.
But I can create a resttest if you want?

msavy (approved):

Issue #2283, opened 2022-08-25
Filter for endpoint '/search/apis' throws errors


BastianGem:
### Apiman Version

3.0

### Apiman Manager Distro

Tomcat

### Apiman Gateway Distro

Vert.x

### Java Version

11

### Operating System

Manjaro

### Are you running Apiman in a container, or on an orchestration platform?

Kubernetes

### Describe the bug

Filtering for numbers with the operators gt, gte, lt, lt or neq will always fail when searching for APIs

Example Body: `{
  "filters": [
    {
      "name": "numPublished",
      "value": "0",
      "operator": "gt"
    }
  ]
}`

Distro: Tomcat
DB: Postgres

### Expected behaviour

We should receive all APIs matching the filters as response.

### Actual behaviour

We receive a StorageException, see below:

`{
  "type": "SystemErrorException",
  "errorCode": -1,
  "message": "io.apiman.manager.api.core.exceptions.StorageException: java.lang.IllegalArgumentException: Parameter value [0] did not match expected type [java.lang.Integer (n/a)]",
  "moreInfoUrl": null,
  "stacktrace": null
}`

### How to Reproduce

Make a request against '/search/apis/ with the following filter 

 `{
  "filters": [
    {
      "name": "numPublished",
      "value": "0",
      "operator": "gt"
    }
  ]
}`

msavy:
Did you check the server logs to get the full exception? Or does it not show one?

PR #2284, opened 2022-08-25
fix(manager-api-jpa): parse filter number into same numeric data type as target field

msavy:
Previously, we were assuming all the fields were `Long`. In some cases they are `Integer` and this causes a data type mismatch with certain target fields.

Fixes: #2283


BastianGem (approved):
LGTM :)

msavy:
Thanks!

Issue #2100, opened 2022-06-09
Allow logging of metrics into file (for scraping into metrics platforms) 


msavy:
## Feature request

In some metrics architectures a scrape-like approach is used whereby the application must write metrics to a local file, and then an agent will scrape and inject into the metrics/logging platform, such as Elasticsearch. 

This has the advantage that the metrics are still potentially available even if ES goes down for an extended maintenance window.

It has the downside that it relies on disk performance and async logging is a must to avoid horrible performance penalties.  

## Acceptance criteria

* Log data to a specified file location (i.e. not remote)
* Performance of normal gateway must not affected (e.g. no material increases in latency)
* Support for rolling etc?

PR #2293, opened 2022-09-01
feat(metrics-es): allow logging metrics to JSON file with `write-to` option

msavy:
To facilitate scrape-based metrics, this commit allows Apiman's ES metrics (as JSON) to be written
to a log file. The specific behaviour is controlled via whatever logging framework is used.

It is possible to write to any of: `remote` (ES server), `log`

The configuration provided in this commit calls the output `metrics.log`, but we the user can change
this, as per their preferences. Async logging is used, where possible, to mitigate the impact on
performance, as this is likely to generate a considerable number of log writes.

Thanks again to Terry, Russel, Joe, Colin, LN, Shivraj, and the rest of the team.

Fixes: #2100


msavy:
Will likely wait for 3.1 to merge this, but let's see.

msavy (commented):
ready

PR #2325, opened 2022-09-27
chore(deps): bump Postgres driver to 42.5.0

msavy:


msavy:
Tested, all seems to work fine.

Issue #2402, opened 2022-12-20
Gateway can not communicate with Keycloak if the KC URL contains a path


volkflo:
### Apiman Version

3

### Apiman Manager Distro

Tomcat

### Apiman Gateway Distro

Vert.x

### Java Version

11

### Operating System

Linux

### Are you running Apiman in a container, or on an orchestration platform?

Kubernetes

### Describe the bug

The current implementation tries to set the path `realms/<your-realm` automatically, based on the configured `auth-server-url`. 
The problem with this is that the `URIBuilder` replaces existing paths that may be needed not communicate to Keycloak.
https://github.com/apiman/apiman/blob/1049b4a9d109fb76db86a4436935b46eef8bc3be/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/api/auth/KeycloakOAuthFactory.java#L188-L190
For example you can have your Keycloak avaiable at `http://host:port/my/custom/path/realms/apiman`

### Expected behaviour

The realm-path should be added instead of a full replace.

### Actual behaviour

The existing path gets replaced by the realm-path.

PR #2403, opened 2022-12-20
fix(gateway): allow additional path in (keycloak) authServerUrl

volkflo:
The previous implementation did not supported URL like "http://localhost:8080/additional/path". The additional path was always replaced.

Closes: #2402


msavy:
Closed via https://github.com/apiman/apiman/commit/71ed74f02c7170e63efc2a99b3742550eeaadec6

PR #2407, opened 2022-12-20
Little UI fix - API Definition

volkflo:
Currently, we always send the "Authorization" header in all apiman applications as this avoids problems if the Keycloak OAuth policy is applied. This results now in the same behavior as the Swagger UI for the Apiman REST API itself and like it is handled in the developer portal.


msavy:
I'm a bit worried this could add a security issue in certain conditions (i.e. accidentally forward an authorization header that is not intended to be sent to the particular backend, which could then expose credentials unintentionally)

PR #2408, opened 2022-12-20
fix(metrics-es): add missing remoteAddr field to EsMetrics

volkflo:
Due to the changes in `2cdace93e0a7a06c6dde9ee1aaf245fb179ec9ba` the `remoteAddr` field was missing in the new class `EsMetricsPayload`. Adding this field again, adds it also in the index.

The field is already part of the index mapping definition.
See:
https://github.com/apiman/apiman/blob/dabb5af66de5a9abc8ed875ce921e2c617e39d4c/gateway/engine/es/src/main/java/io/apiman/gateway/engine/es/EsMetrics.java#L259


msavy:
I wonder if this is just in your downstream distro and you added something extra to the metrics? This fails with a compilation error that suggests that might be the case.

msavy:
Thanks for your comments.

The issue would be we'd probably need to add it to all the other metrics impls including JDBC which might be a breaking change as I don't think we have any ability to do migrations yet for implementations in that area.

I'll have a think about what to do.

msavy:
Consider adding `remoteAddr` to Apiman metrics

PR #2464, opened 2023-01-20
test(gateway-test): refactor flaky transfer-quota test by using day granularity

msavy:
Use daily time granularity to avoid issue where test could fail due to coincidental rollover into new hour.

Add some data to the request to make the test more realistic.

### DIFF
--- a/gateway/test/src/test/resources/test-plan-data/policies/transfer-quota/009-probe-ip-transfer-quota.resttest
+++ b/gateway/test/src/test/resources/test-plan-data/policies/transfer-quota/009-probe-ip-transfer-quota.resttest
@@ -15,13 +15,12 @@ Content-Type: application/json
 {
   "TransferQuotaProbeResponse": {
     "config": {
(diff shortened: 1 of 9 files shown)


gateway/test/src/test/resources/test-plan-data/policies/transfer-quota/009-probe-ip-transfer-quota.resttest
```diff
@@ -15,14 +15,13 @@ Content-Type: application/json
...
     "config": {
-      "limit": 100,
+      "limit": 355,
       "granularity": "Ip",
-      "period": "Hour"
+      "period": "Day"
     },
     "status": {
-      "accepted": true,
-      "remaining": 100
```
msavy, line 24: I've removed remaining because it seems to differ slightly by platform and we don't have a great way of putting in numeric ranges in currently with the RestTest suite. Maybe something for future.

PR #2474, opened 2023-02-16
bug fixing, allowedIssuers validation

ronimhd:
bug fixing, if allowedIssuers contain issuerEndpoint, validation should not give error


msavy:
I need to rebase this branch, so will close this here but push and credit you

PR #2481, opened 2023-03-06
fix(ui): avoid undefined errors

volkflo:
If one switches between ip policies the ip-list could be undefined. Moving the check into the validate method fixes the problem.

### DIFF
--- a/manager/ui/war/pom.xml
+++ b/manager/ui/war/pom.xml
@@ -172,7 +172,7 @@
             </goals>
             <phase>generate-resources</phase>
             <configuration>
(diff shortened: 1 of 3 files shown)


manager/ui/war/pom.xml
```diff
@@ -172,7 +172,7 @@
             </goals>
             <phase>generate-resources</phase>
             <configuration>
-              <arguments>--mode production --stats verbose</arguments>
```
msavy, line 175: I assume this is unrelated to the bug fix?

PR #2510, opened 2023-04-19
fix(gateway-vertx): do not include null/empty elements in Keycloak di…

msavy:
…scovery URI


msavy (commented):
Add author to changelog

msavy:
Fixes #2513

PR #2511, opened 2023-04-19
fix(distro-ddl):  fix multiple MSSQL DDL issues

msavy:
∙ `hibernate_sequence` creation was limited to H2 and Postgres only for some reason.

∙ `GO` statement separators in gateway DDL did not play nicely with Apiman's Gateway DDL runner.

∙ Explicit database name was used in Gateway DDL which caused issues if using non-default DB name.

∙ Use custom JSON type to make handling of JSON in MSSQL easier when also using Hibernate.


msavy:
Fixes #2512

PR #2518, opened 2023-04-21
fix(manager-api-jdbc): quote plan keyword JdbcMetricsAccessor for  MSSQL if `hibernate.auto_quote_keyword` set

msavy:


msavy:
Closed after pushing directly to master (I changed the commit messages slightly).

Issue #2542, opened 2023-06-20
API definition rewrite cache does not work


volkflo:
### Apiman Version

3

### Apiman Manager Distro

Tomcat

### Apiman Gateway Distro

Vert.x

### Java Version

11

### Operating System

Arch

### Are you running Apiman in a container, or on an orchestration platform?

Docker / Podman

### Describe the bug

If the API is published and you try to load the API definition the basePath / host is rewritten to the actual api endpoint on the gateway.
After rewriting the definition is put into a cache.
However items are not taken out of the cache as the cache key object does not implement equals/hashCode and therefore is always a new one for each request.

### Expected behaviour

Definitions are correctly pushed into cache and taken out of the cache.

### Actual behaviour

Definitions are put into the cache but not resolved from cache.

PR #2543, opened 2023-06-20
refactor(manager-api): change cacheKey for api definition rewrite cache

volkflo:
Use `modifiedOn` as key to make sure that we refresh the cache once the definition is updated. Otherwise definition changes are not directly available after an update. Also this fixes caching as the cacheKey class did not override equals/hashCode.

Closes: #2542

See: https://github.com/ben-manes/caffeine/discussions/603

### DIFF
--- a/manager/api/rest-impl/src/main/java/io/apiman/manager/api/schema/SchemaRewriterService.java
+++ b/manager/api/rest-impl/src/main/java/io/apiman/manager/api/schema/SchemaRewriterService.java
@@ -20,7 +20,9 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.nio.charset.StandardCharsets;
 import java.time.Duration;
+import java.util.Date;
 import java.util.Map;
+import java.util.Objects;
 import java.util.StringJoiner;
 import javax.inject.Inject;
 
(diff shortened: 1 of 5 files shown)


manager/api/rest-impl/src/main/java/io/apiman/manager/api/schema/SchemaRewriterService.java
```diff
@@ -124,27 +126,40 @@ public String rewrite(ProviderContext providerCtx, InputStream is, ApiDefinition
...
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (o == null || getClass() != o.getClass()) return false;
+            SchemaCacheKey that = (SchemaCacheKey) o;
+            return apiVersionId == that.apiVersionId && Objects.equals(modifiedOn, that.modifiedOn);
+        }
+
+        @Override
+        public int hashCode() {
```
msavy, line 161: Whoops, bad oversight. This is why Java Records are a blessing.

msavy (commented):
I think this is probably a bug fix rather than a refactor. What do you think?

volkflo:
Ah yes, makes more sense: You can change the commit message if you want to :)

msavy:
Thanks, Florian. These look very reasonable.

manager/api/rest-impl/src/main/java/io/apiman/manager/api/schema/SchemaRewriterService.java
```diff
@@ -124,27 +126,40 @@ public String rewrite(ProviderContext providerCtx, InputStream is, ApiDefinition
 
     private static class SchemaCacheKey {
         private final long apiVersionId;
-        private final ApiDefinitionType apiDefinitionType;
```
msavy, line 127: Is the `ApiDefinitionType` not also needed?

msavy: Not saying it _is_, but just want to understand why it was removed rather than 3-part composite key.

msavy: I wrote this a while ago, so may need to refresh my understanding of it.

msavy: Okay, makes sense.

msavy (approved):

Issue #2557, opened 2023-07-04
Time-Restricted-Access-Policy uses gateways local time zone instead of UTC


volkflo:
### Apiman Version

3

### Apiman Manager Distro

Tomcat

### Apiman Gateway Distro

Vert.x

### Java Version

11

### Operating System

Manjaro

### Are you running Apiman in a container, or on an orchestration platform?

No (Bare Metal)

### Describe the bug

The time restricted access policy is configured with a UTC string like this:

```
{
     "timeStart": "06:00:00",
     "timeEnd": "14:00:00",
}
```

The problem with this is that this is parsed to a `java.util.Date` field (which does not have a time zone).
During deserialization with jackson the Date is then created in local time then.

For example on my test gateway  (Europe/Berlin - CET) the configuration from above gets parsed as:
![image](https://github.com/apiman/apiman/assets/13332383/076c3228-d7c1-463d-aba3-be057a7160e6)

The problem with that is that the config does not match the time range that should be blocked.

### Expected behaviour

The Gateway treats the configuration always as UTC.

### Actual behaviour

The policy only works if the gateway is running/set to a timezone without offset (UTC)

### Anything else?

I will provide a PR soon, that fixes that behavior.

msavy:
In future versions of Apiman I'm trying to replace them all

TZ should probably be configurable

volkflo:
Yeah makes sense, seems that would be a good change and make things more predictable.
I just fixed this now for this one policy and replaced all joda stuff with OffsetDateTime (as this seems the most flexible one).

PR #2558, opened 2023-07-04
fix: ensure use of UTC for the time-restricted-access-policy

volkflo:
During the previous deserialization of `java.util.Date` the config was parsed by jackson to local time of the gateway instead of UTC. Date and joda-time are replaced with java default `OffsetDateTime` now, which also allows us to be more flexible in future. Old configuration are still compatible. Test is added back again and fixed.

Closes: #2557 

## Checklist

- [x] I have added a changelog entry.
- [X] I have tested this change manually, as well as as passing tests.

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -13,6 +13,9 @@ All notable changes to Apiman will be documented here (as of Apiman 3).
 
 === Fixed
 
+* [gateway]: Use UTC for Time Restricted-Access-Policy. The Gateway used the configured local (server) time instead of UTC while applying the rules. This could have lead to a mismatch between configuration (UTC) and execution.
+By https://github.com/volkflo[Florian Volk (@volkflo)^].
+
 // end::3.2.0-SNAPSHOT[]
 
 // tag::3.1.2.Final[]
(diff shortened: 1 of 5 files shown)


msavy (commented):
I think this looks good and I think I understand what you are trying to achieve, can you please complete the checklist though 🙏

* Will this break any existing behaviour (e.g. any potential reliance on "wrong" behaviour?)? 
* Do we always transform to UTC in the UI? Trying to remember how we handle local date-time requirements.
* Note for self: I also need to fix all the time-related problems for Apiman 4, which is quite a big headache so far, tbh.

volkflo:
Sorry, the changelog was new for me, did not saw that on first hand

> * Will this break any existing behaviour (e.g. any potential reliance on "wrong" behaviour?)?

Yes and no, it depends on how you view on the change I guess. 
If you have running the gateway in a different TZ than UTC the rules are applied in local time.
With this fix they would be executed in UTC now.
However the configuration was always done in UTC, so from that point of view we just fix a bug.

> * Do we always transform to UTC in the UI? Trying to remember how we handle local date-time requirements.

In the UI you configure it in your local time which is than converted to UTC (which makes config easier).
I made a contribution a while ago that shows you what you configured in which TZ.
image

msavy:
I think the fix is correct then 👍

PR #2570, opened 2023-07-19
fix(gateway-vertx): use cache for non-shared JDBC client

### LINKED ISSUE #2569: Vert.x Basic Authentication JDBC - Driver missing and pooling problems
### Apiman Version

3.X

### Apiman Manager Distro

Tomcat

### Apiman Gateway Distro

Vert.x

### Java Version

11

### Operating System

Manjaro

### Are you running Apiman in a container, or on an orchestration platform?

No (Bare Metal)

### Describe the bug

Currently the bundled vert.x gatway jar does not contain any jdbc driver.
That makes the built in Basic Auth Policy with JDCB unusable.

In addition if you add the drives the implementation creates a new client for each request which leads to connection pooling problems after a while.

### Expected behaviour

- Include JDBC Driver as we do it for wildfly (can be discussed to keep size smaller)
- Clients are cached

### Actual behaviour

_No response_

### How to Reproduce

1. Create a public API
2. Attach Basic Auth Policy with a JDBC source
3. Publish API
4. Make request with basic auth

### Relevant log output

_No response_

### Anything else?

_No response_

volkflo:
## Description

Adding the default set of apiman jdbc drivers to vert.x. In addition cache non-shared clients (e.g. clients for basic-auth-policy) instead of creating a new once per request.

This is maybe not the perfect solution but I did not wanted to refactor the whole client logic as this may introduce additional bugs/breaking stuff.

Fixes #2569

## Checklist

- [ ] I have added a changelog entry.
- [ ] If I've added a new feature/enhancement, I have added documentation for it.
- [ ] I have tested this change manually, as well as as passing tests.

### DIFF
--- a/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/components/jdbc/VertxJdbcClientImpl.java
+++ b/gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/components/jdbc/VertxJdbcClientImpl.java
@@ -15,6 +15,9 @@
  */
 package io.apiman.gateway.platforms.vertx3.components.jdbc;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+import com.github.benmanes.caffeine.cache.RemovalCause;
 import io.apiman.gateway.engine.async.AsyncResultImpl;
 import io.apiman.gateway.engine.async.IAsyncResultHandler;
 import io.apiman.gateway.engine.components.jdbc.IJdbcClient;
@@ -25,24 +28,27 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.ext.jdbc.JDBCClient;
 import io.vertx.ext.jdbc.spi.impl.HikariCPDataSourceProvider;
 
-import java.util.Map.Entry;
-
 import javax.sql.DataSource;
+import java.util.Map.Entry;
 
 /**
  * @author Marc Savy {@literal <msavy@redhat.com>}
  */
 public class VertxJdbcClientImpl implements IJdbcClient {
-    //TODO cache jdbcClients && json configs? LinkedHashMap + removeEldestEntry?
-    //private static Map<>
-    private JDBCClient jdbcClient;
+    private static final Cache<String, JDBCClient> cache = Caffeine.newBuilder()
+            .maximumSize(20)
+            .evictionListener((String key, JDBCClient client, RemovalCause cause) -> closeOnEviction(client))
+            .build();
+
+    private final JDBCClient jdbcClient;
 
     public VertxJdbcClientImpl(Vertx vertx, String dsName, JdbcOptionsBean config) {
         jdbcClient = JDBCClient.createShared(vertx, parseConfig(config), dsName);
     }
 
     public VertxJdbcClientImpl(Vertx vertx, JdbcOptionsBean config) {
-        jdbcClient = JDBCClient.createNonShared(vertx, parseConfig(config));
+        // currently only cache non-shared clients as shared clients already share the same ds/pool
+        jdbcClient = cache.get(config.getJdbcUrl() + "-" + config.getUsername(), k -> JDBCClient.create(vertx, parseConfig(config)));
     }
 
     public VertxJdbcClientImpl(Vertx vertx, DataSource ds) {
@@ -62,7 +68,7 @@ public class VertxJdbcClientImpl implements IJdbcClient {
 
     /**
      * Translate our abstracted {@link JdbcOptionsBean} into a Vert.x-specific config.
-     *
+     * <p>
      * We are assuming that the user is using HikariCP.
      */
     @SuppressWarnings("nls")
@@ -94,5 +100,12 @@ public class VertxJdbcClientImpl implements IJdbcClient {
         if (value == null)
             return;
         jsonObject.put(key, value);
+    }
+
+    // Explicitly close the client when it gets evicted from the cache
+    private static void closeOnEviction(JDBCClient client) {
+        if (client != null) {
+            client.close();
+        }
     }
 }
--- a/gateway/platforms/vertx3/vertx3/pom.xml
+++ b/gateway/platforms/vertx3/vertx3/pom.xml
@@ -176,6 +176,28 @@
       <groupId>org.jboss.spec.javax.ws.rs</groupId>
       <artifactId>jboss-jaxrs-api_2.1_spec</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.github.ben-manes.caffeine</groupId>
+      <artifactId>caffeine</artifactId>
+    </dependency>
+
+    <!-- DB drivers -->
+    <dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.mysql</groupId>
+      <artifactId>mysql-connector-j</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.oracle.database.jdbc</groupId>
+      <artifactId>ojdbc11</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.microsoft.sqlserver</groupId>
+      <artifactId>mssql-jdbc</artifactId>
+    </dependency>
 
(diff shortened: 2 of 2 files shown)


msavy:
Thanks. I considered this approach but discounted it for a few reasons:

1. It increases the size;
2. It forces a particular JDBC driver version which is difficult for the users to replace/substitute;
3. It includes drivers which have licenses which are not compatible with the license of the main project;

I've backported a solution for my 3.x customers that allows them to include the JDBC driver externally via a workaround and some additional code, but I am not adding new features to 3.x upstream at this point.

This will be fixed in the upstream properly with Apiman 4 (probably).

The cache change is useful though by itself.

volkflo:
Thanks for the feedback Marc!
I already thought that the drivers are problematic for the upstream, I just copied them over from the wildfly pom to test everything.

The main problem we encountered was the client that was created again and again.
So the cache was useful in our case. I just tried to use the same libs you used in other places for this.

msavy:
> Thanks for the feedback Marc! I already thought that the drivers are problematic for the upstream, I just copied them over from the wildfly pom to test everything.
> 
> The main problem we encountered was the client that was created again and again. So the cache was useful in our case. I just tried to use the same libs you used in other places for this.

Yeah, the way we get away with it is that we never *distribute* the project or code with those license-incompatible drivers. They are referenced only for test or development purposes such as generating DDL, etc. 

I'm probably going to drop some of them for upstream anyway, which should hopefully fix that issue entirely.

Yet more things for me to consider in Apiman 4 I suppose. I'm not getting to spend as much time on it as I would like, but still trying to push ahead.

msavy:
OK, let's split this topic.

I'll commit the cache enhancement separately, and then we can think separately about whether we should include any drivers in 3.x or just wait for Apiman 4 to avoid complicating things.

gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/components/jdbc/VertxJdbcClientImpl.java
```diff
@@ -25,24 +28,27 @@
...
 
 /**
  * @author Marc Savy {@literal <msavy@redhat.com>}
  */
 public class VertxJdbcClientImpl implements IJdbcClient {
-    //TODO cache jdbcClients && json configs? LinkedHashMap + removeEldestEntry?
-    //private static Map<>
-    private JDBCClient jdbcClient;
+    private static final Cache<String, JDBCClient> cache = Caffeine.newBuilder()
+            .maximumSize(20)
```
msavy, line 39: Is 20 just a magic number?

msavy:
I've thought of a potential solution that's even better than the workaround I use for customers to allow an external JDBC driver, so I'll experiment with that later when I have some time 🤞. Posting this as a reminder to myself.

volkflo:
> OK, let's split this topic.
> 
> I'll commit the cache enhancement separately, and then we can think separately about whether we should include any drivers in 3.x or just wait for Apiman 4 to avoid complicating things.

I think that is a good idea.  I can remove the drives from my commit if you want?

msavy:
I'm also trying to get rid of the old ES libraries that were dragged in during a contribution a few years ago which has massively bulked out the gateway. That's a WIP for Apiman 4 (usual caveats apply for time and money, etc).

gateway/platforms/vertx3/vertx3/src/main/java/io/apiman/gateway/platforms/vertx3/components/jdbc/VertxJdbcClientImpl.java
volkflo: Yes, I thought it would maybe make scene to make this configurable but not sure if there is really a need for it.
Currently it would cache 20 clients (20 different db configs, I am not sure if we really have the need for more.)
If you would use the same db table for different basic-auth-policy configs it would be always the same client.

volkflo:
> Whichever you prefer, sounds good to me pray

Okay I will do it and directly add the changelog here as well :)

msavy:
Okay, related thought... Is there any reason _not_ to use a Vert.x shared connection pool any more? Can we generate a key for the Vert.x shared connection instead of caching our own? 

Thinking out loud here.

msavy:
For example, if we are not sure we can reliably create a key that will guarantee we don't mix up the wrong clients (which could be a problem with this enhancement anyway I guess).

msavy:
Also, should users of the JDBC client not be closing the connection themselves? Do we not make that available to them.

volkflo:
Mhh you may be right about that, my vert.x knowledge is not that deep.
We could do it in the same way like we do it here:
```
jdbcClient = JDBCClient.createShared(vertx, parseConfig(config), dsName);
```
I was not totally sure if we really need a new instance of the client all the time:
> Subsequent calls will return a new client instance that uses the same data source, so the configuration won’t be used.
https://vertx.io/docs/3.9.16/vertx-jdbc-client/java/#_using_default_shared_data_source

volkflo:
> Also, should users of the JDBC client not be closing the connection themselves? Do we not make that available to them.

Currently I cannot see in the logs that a client is close.
Only when the `close` is called I see the following log:
```
 HikariDataSource - HikariPool-1 - Shutdown completed.
```

volkflo:
Not sure about the CI:
> Error:  Errors: 
Error:    ExportTest.export-testPlan.xml / Export (2) Download »  com.fasterxml.jackson....
[INFO] 
Error:  Tests run: 778, Failures: 0, Errors: 1, Skipped: 0

This seems to not relate with the fix

msavy:
Yeah, this requires a careful look because if you get the cache key wrong, then you may end up mixing together unrelated SQL calls and that will break things in extremely confusing ways. 

I recall I avoided implementing caching at the time for that reason. That's why the commen references "jsonConfigs" as I was trying to figure out what needed to be included in the cache key to guarantee uniqueness.

It might actually be easier to change the code calling this impl, not sure...

I will try to look soon.

volkflo:
I checked that as well and it seems to work as well.
The only things is that we would also need to provide a unique key (even for the shared).

With a dsName it uses a default name (which than would return the same client for all configs) 
```
static JDBCClient createShared(Vertx vertx, JsonObject config) {
        return new JDBCClientImpl(vertx, config, "DEFAULT_DS");
}
```

msavy:
Question - why is the calling code not setting a `dsName` of their own, to ensure it is shared. From a quick review, it seems we're just doing that out of precaution that we're not sure what the correct cache key is (same problem again!).

Couple of thoughts: 

Wonder if we could just set a `hashcode` and `equals` function in `JdbcOptionsBean` and use that as the safest possible cache key, and always use a shared client? We'd need to ensure we account for unlikely hashcode collisions in our naming I suppose, but we could do something like...

```
Map<JdbcOptionsBean, JdbcConnection>  CACHE... 

JDBCClient.createShared(<usual options>, "apiman-" + RandomUUID); <-- we'll manage the mapping ourselves via the cache/map above, so key doesn't matter and we can avoid collision issues
```

In worst case it would end up creating more pools than needed if people pass in lots of different configuration permutations, but it should avoid ever accidentally getting caught out (e.g. user passes an extra option that means it's no longer compatible with the simple `jdbcUrl+username` cache key scheme).

Thoughts?

volkflo:
I think if we would use the complete config as cache key then we would´t need a shared client anyway? 
Then we could use a normal client (which ensures its created with a unique id).

Only question is what happens with all the clients in the map? There would be no limit of different clients then?
Not sure if this could cause problems? 

---

I think if we would create a unique dsName already here, it would be solved as well:
```
  public IJdbcClient createStandalone(JdbcOptionsBean config) {
        return new VertxJdbcClientImpl(vertx, config);
    } 
```

msavy:
How would it work with your proposal. I don't see how we would otherwise
ensure there's no cache collision without something clumsy like a building
a string every time.

msavy:
I guess the problem is twofold:

The Apiman code is using non-shared in places it should probably use shared.

And are we breaking the non-shared case by some cache mappings (I think it
should be ok, but not sure).

volkflo:
Ah, I did not explained it in a good way.
I thought about a combination.
Maybe its not a good idea just thinking loud :D

1. implement hashCode/equlas
2. use the hashCode as unique dsName

So we would not create a shared client which is still unique and vertx would handle it for us.
Then we may do not even need an additional Map<config,clients>

PR #2655, opened 2024-10-07
feat: support key lookup based on kid in jwt token header

volkflo:
JWK(s) endpoints often perform key rotation, so the cryptographic key should be looked up based on the `kid` inside the JWT header.

To maintain backward compatibility, the policy first looks up the `kid` from the configuration. If no `kid` is specified in the policy configuration, the policy then checks for the `kid` in the token header and fetches the JWK(s) to obtain the correct public key.

See also: https://github.com/jwtk/jjwt#dynamic-key-lookup


msavy:
Thanks for the contribution! I'll review soon and try to merge this with a
similar change I have staged for Apiman 4.

msavy:
Will leave this PR open if that's ok. So I will remember to review and merge anything I may have missed.

msavy:
Sorry, I'm on mobile right now and can't figure out how to do it from the app 😆😂. I'll reopen once I get to my destination, which shouldn't be too long. 

Thanks for your patience.

